PR #27, opened 2018-05-18
WIP: Activitystreams crate

### LINKED ISSUE #17: Use the activitystreams crate
This crate looks amazing, and I'm sure we would benefit from using it: https://github.com/asonix/activitystreams

BaptisteGelez:
Still a lot of `// TODO` to resolve

Closes #17


asonix:
Do you feel like using this crate is better than what you were doing before?

Is there anything that it doesn't do that you wish it did?

BaptisteGelez:
@asonix I think my code is more idiomatic now, and the way we are serializing/deserializing activities is really cleaner. One thing I would have liked to completely avoid, is to manipulate the JSON object directly (as I'm doing here, for instance, https://github.com/Plume-org/Plume/pull/27/files#diff-a244a0dbf50367a4f27f1f20df813475R48 because I didn't found any clean solution). But it globally nice to use, even if it is hard to have a nice API for ActivityStreams in Rust.

asonix:
Yeah, custom properties is something that I've had trouble with figuring out. My method of doing it now is through the CustomLink and CustomObject types

You can do

CustomObject::new(your_base_object, your_extra_properties) and it will implement Object if your_base_object implements Object, actor if your_base_object implements Actor, etc.

BaptisteGelez:
Nice! Thanks for your help.

Issue #25, opened 2018-05-15
Fallback to username if no display name is available


BaptisteGelez:
For instance, nothing is displayed here: https://baptiste.gelez.xyz/~/MonBlog/plume-development-comments-are-coming/#comment-54

JosephKiranBabu:
It is currently possible to create accounts without any usernames. The system seems to be already falling back to usernames when no display name is set.

I was able to reproduce this by creating an account without a username. I only gave email and password in the registration form. Then I was able to create nameless comments just like those in the demo instance. I think the correct solution here is to disallow creating accounts without usernames.

BaptisteGelez:
> It is currently possible to create accounts without any usernames.

Yes, but it is a different problem.

> The system seems to be already falling back to usernames when no display name is set.

I can confirm it's not. :sweat_smile:

BaptisteGelez:
Closed in #50

PR #50, opened 2018-06-11
Fallback to username if no display name is available

stephenburgess8:
Addressing #25 

I also changed it so that if a Post's author doesn't have a display name set, the Post will show the author's username.

### DIFF
--- a/templates/posts/details.html.tera
+++ b/templates/posts/details.html.tera
@@ -9,8 +9,13 @@
 {% endblock header %}
 
 {% block content %}
+{% if author.display_name %}
+    {% set name = author.display_name %}
+{% else %}
+    {% set name = author.username %}
+{% endif %}
     <p>
-        <b>Written by <a href="/@/{{ author.fqn }}/">{{ author.display_name }}</a></b>
+        <b>Written by <a href="/@/{{ author.fqn }}/">{{name }}</a></b>
         &mdash;
         <span>{{ date | date(format="%B %e, %Y") }}</span>
     </p>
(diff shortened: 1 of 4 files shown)


templates/posts/details.html.tera
```diff
@@ -49,8 +54,13 @@
         <h2>Comments</h2>
         <div class="list">
             {% for comment in comments %}
+                {% if comment.author.display_name %}
+                    {% set commentAuthorName = comment.author.display_name %}
```
BaptisteGelez: I would prefer camel_case for variable names, e.g `comment_author_name`

templates/posts/details.html.tera
```diff
@@ -49,8 +54,13 @@
...
+                {% else %}
+                    {% set commentAuthorName = comment.author.username %}
+                {% endif %}
                 <div class="card" id="comment-{{ comment.id }}">
-                    <b>{{ comment.author.display_name }}</b>
+                    <b>{{ name }}</b>
```
BaptisteGelez: Shouldn't it be `comment_author_name`?

stephenburgess8:
Thanks for the feedback. I usually default to camel case for naming anything that is not on the persistence layer, but I don't mind switching to snake here at all.

I have tested everything except for followers, as I haven't gotten around to creating dummy accounts locally to create followers. One thing to notice is that in `user/details`, if `author.display_name` isn't set, the fallback is `user.username`, because `author.username` was not a valid key.

BaptisteGelez:
@stephenburgess8 I think it's normal `author.display_name` is not working on `user/details`, since I don't think `author` is defined for this page, the variable containing informations about the profile we are viewing is `user`. So, I think you should just use `user` everywhere on this page.

stephenburgess8:
Oops, of course what you say makes perfect sense. Thanks for the help.

templates/users/details.html.tera
```diff
@@ -2,13 +2,17 @@
 {% import "macros" as macros %}
 
 {% block title %}
-{{ user.display_name }}
+{% if author.display_name %}
```
BaptisteGelez: It feels a bit strange to have this `if` in the `title` block, it would be perfect if you moved it outside. Else, everything seems to be working just fine, great job!

templates/users/followers.html.tera
```diff
@@ -1,13 +1,19 @@
 {% extends "base" %}
 
 {% block title %}
-{{ user.display_name }}'s Followers
+{% if user.display_name %}
```
BaptisteGelez: Same here: it would be great if you could move this `if` outside of the `title` block.

BaptisteGelez:
Thanks a lot!

PR #60, opened 2018-06-17
I18n

### LINKED ISSUE #58: i18n support
Just like in title, it might be challenging, because I haven’t seen any good i18n crate for web, but it’s definitely possible. My Rust knowledge is not enough to start working on it. I will translate the project to Polish when it’s done.

BaptisteGelez:
Here is the workflow I came with:
- When we want to add a new locale, just list it in `po/LINGUAS`
- When we want to translate something:
  - Use the `_` or `_n` filters in the Tera template
  - Add the corresponding string to `po/plume.pot`
- Then when you run `cargo build` (or `cargo run`), it:
  - Updates your translation files (`.po` ones) with new strings, or create them from the POT if they don't exist
  - Compile them to `.mo` files
- On the Rocket side:
  - When launching, tell gettext where our `.mo` files are
  - For each request, fetch the requested locale from the `Accept-Language` header, and tell gettext to use it (with a dirty hack)

TODO:

- [x] Make the interface actually translatable (by replacing untranslatable strings with calls to `_` and `_n`)
- [x] Write an internationalization guide
- [x] Compile translation when launching instead of at compile-time, to avoid useless builds (and make it easier to integrate with other projects in the future)
- [x] Put the `i18n` module in its own crate?

Fixes #58


BaptisteGelez:
I made a crate for this, so that other Rocket-based projects can re-use it: https://crates.io/crates/rocket_i18n

templates/instance/index.html.tera
```diff
@@ -8,7 +8,7 @@
 {% block content %}
     <h1>{{ "Welcome on {{ instance_name }}" | _(instance_name=instance.name) }}</h1>
 
-    <h2>Latest articles</h2>
+    <h2>{{ "Latest articles" }}</h2>
```
fdb-hiroshima: I think a `_` is missing here

fdb-hiroshima:
In-code strings (I'm thinking strings used to inform about login reason like here https://github.com/Plume-org/Plume/blob/5ae9cb3490a40b3e48b53b56d212035bc26695cc/src/routes/blogs.rs#L59 ) are missing i think

PR #66, opened 2018-06-19
Setup script

### LINKED ISSUE #47: /configure public accessible
[/configure](http://baptiste.gelez.xyz/configure) is currently publicly accessible. This could be a security issue.

BaptisteGelez:
When we run Plume, check that the database is ready, and that the instance have been created. If it is not, create the database, run migrations, and ask a few questions to set up the instance and create the admin.

Fix #47


.env
```diff
@@ -1,2 +1,6 @@
 
-BASE_URL=plu.me
\ No newline at end of file
+BASE_URL=plu.me
+ROCKET_SECRET_KEY=7IdEGAf7EL7ehPjLNncCcE+2B0MTiywxDzr8KklU/VA=
```
fdb-hiroshima: did you actually wanted to commit a secret key?

src/setup.rs
```diff
@@ -0,0 +1,261 @@
...
+                .arg("plume")
+                .status()
+                .map(|s| {
+                    if s.success() {
+                        println!("{}\n", "  ✔️ Done".green());
+                    }
+                })
+                .expect("Couldn't create new user");
+
+            println!("{}\n", "About to create a new PostgreSQL table named 'plume'".blue());
```
fdb-hiroshima: I think it's not a table but a database

src/setup.rs
```diff
@@ -0,0 +1,261 @@
...
+}
+
+fn complete_setup(conn: DbConn) {
+    quick_setup(conn);
+
+    println!("\nOn which port should Plume listen? (default: 7878)");
+    let port = read_line_or("7878");
+    write_to_dotenv("ROCKET_PORT", port);
+
+    println!("\nOn which port should Plume listen? (default: 0.0.0.0)");
```
fdb-hiroshima: Shouldn't it be on which ip?

BaptisteGelez: (oops, forgot to change that after having copy/pasted the previous lines)

Issue #68, opened 2018-06-20
Always logged out on error pages


BaptisteGelez:
On error pages, you always have the links to login or create an account, even if you are already logged in.

BaptisteGelez:
Fixed in #71

PR #71, opened 2018-06-21
solve #68

Zanfib:
extract user information from the request object in the error routes, work well on direct 404 and 500 errors but not with the macro may_fail in src/routes/mod.rs


BaptisteGelez:
Thanks for this PR! To fix your issue with the macro, you may try to changes those lines https://github.com/Plume-org/Plume/blob/master/src/routes/mod.rs#L12-14 to:

```rust
Template::render(concat!("errors/", $template), json!({
    "error_message": $msg,
    "account": user
}))
```

But then you should be sure that there always is a variable called `user` containing the logged in user before using `may_fail!`.

If you prefer not to touch the macro, I can also merge this branch as is it is, and fix this part later. :slightly_smiling_face:

Zanfib:
I've find another problem that impact the macro usage, in the "user/edit" route the user name isn't found by the route because the template have a malformed link generation : the url point to /@/edit in place of /@/<name>/edit
If I solve this point the macro is OK I think, I need to investigate and experiment a little bit more ;)

templates/users/header.tera
```diff
@@ -11,7 +11,7 @@
     </h1>
 
     {% if is_self %}
-        <a href="edit" class="button inline-block">{{ "Edit your profile" | _ }}</a>
+        <a href="{{ name }}/edit" class="button inline-block">{{ "Edit your profile" | _ }}</a>
```
BaptisteGelez: This link won't always be valid: if you are on `/@/my_user_name/` instead of `/@/my_user_name` it will be broken. It's always the same issue with trailing slashes. The solution here would be to have an absolute link, like `/@/{{ user.fqn }}/edit/`.

Zanfib: I see the point here, I will test this possibility but I see another point for the future, if you have Plume behind a reverse proxy and in a subdirectory of a bigger web site you have another problem with absolute url i think, or Rocket can manage this case ?

BaptisteGelez: I think it would probably require to modify a bit the code (change the Rocket mount point in `main.rs`), but it should be possible.

Zanfib:
finally got it :) (i think ... if i'm completely wrong reject my work !!)

PR #104, opened 2018-06-30
New forms style

Madeorsk:
Login example on phones and register example on a computer.
Mobile
Computer


BaptisteGelez:
Looking at your screens, it looks really nicer than before. My only concern would  be about the centered text in inputs, since the text is moving in a quite strange way when typing like that. But I still should give it a live try.

BaptisteGelez:
I'm going to merge this and make a few changes after.

PR #159, opened 2018-07-21
Add a link to the blog on article cards

### LINKED ISSUE #134: Add a link to the blog on article cards

Ilphrin:
This should fix #134, even though I am not sure it is visually ok :/ is there any UI prototype? ^^

### DIFF
--- a/plume-models/src/posts.rs
+++ b/plume-models/src/posts.rs
@@ -187,8 +187,7 @@ impl Post {
             "author": self.get_authors(conn)[0].to_json(conn),
             "url": format!("/~/{}/{}/", blog.actor_id, self.slug),
             "date": self.creation_date.timestamp(),
-            "blog": blog,
-            "blog_url": format!("/~/{}", blog.actor_id)
+            "blog": blog.to_json(conn)
         })
     }
 
--- a/plume-models/src/posts.rs
+++ b/plume-models/src/posts.rs
@@ -181,11 +181,14 @@ impl Post {
     }
 
     pub fn to_json(&self, conn: &PgConnection) -> serde_json::Value {
+        let blog = self.get_blog(conn);
         json!({
             "post": self,
             "author": self.get_authors(conn)[0].to_json(conn),
-            "url": format!("/~/{}/{}/", self.get_blog(conn).actor_id, self.slug),
-            "date": self.creation_date.timestamp()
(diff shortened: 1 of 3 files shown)


plume-models/src/posts.rs
```diff
@@ -181,11 +181,14 @@ impl Post {
...
+        let blog = self.get_blog(conn);
         json!({
             "post": self,
             "author": self.get_authors(conn)[0].to_json(conn),
-            "url": format!("/~/{}/{}/", self.get_blog(conn).actor_id, self.slug),
-            "date": self.creation_date.timestamp()
+            "url": format!("/~/{}/{}/", blog.actor_id, self.slug),
+            "date": self.creation_date.timestamp(),
+            "blog": blog,
+            "blog_url": format!("/~/{}", blog.actor_id)
```
BaptisteGelez: This won't work with federated blogs, since `actor_id` only contains the part before the `@`. The best solution would be be to use `blog.to_json` for the `"blog"` field above, and then use the `fqn` field in the view to build the URL.

(BTW, I'm seeing that the `url` field is wrong too, but that's another issue)

templates/macros.html.tera
```diff
@@ -18,6 +18,7 @@
                 link_4="</a>")
             }}
             ⋅ {{ article.date | date(format="%B %e") }}
+            ⋅ <a href="{{article.blog_url}}">{{article.blog.title}}</a>
```
BaptisteGelez: About the code style: if you could add spaces between the braces and the expression they contain, it would be more readable (and consistent with what was done before).

plume-models/src/posts.rs
Ilphrin: The problem is that Blog doesn't have an implementation of to_json :/ (and if I try using the macro `json!`, it works but I don't have the field fqn)

BaptisteGelez: Ok, your solution looks good then. And we can always add the `Blog::to_json` function in the future to make things simpler.

Ilphrin: Yep sure ^^
Then, how can I get the url of the corresponding instance? As I don't have access to a fqn property

Ilphrin: I can't also take some time to write a basic method for `Blog::to_json` if you wish ^^

PR #168, opened 2018-08-01
Update German translation

bitkeks:
Note that the split in [#L431](https://github.com/bitkeks/Plume/blob/93941cdff86450f517ead0d14d61f2a50d8f476d/po/de.po#L431) is different in German, since the sentence is split in two `msgid`s and is structured differently when translated :wink:


BaptisteGelez:
Thank you!

PR #174, opened 2018-08-18
Rename 'reshare' to 'boost' in templates and PO

### LINKED ISSUE #74: Rename reshares
I don't like the term "reshare". I understand that the idea behind this word is to say that we share a second time, but I think that a second share should still be named a "share".

We use the word "Publish" in the button to create a new blog post, why not using this word again and rename "reshare" to "republish"?

bitkeks:
For now it might be enough to only reword 'reshare' to 'boost' in the
frontend. The code can continue to use 'reshare'.

Includes the most recent German translation.

Fixes #74


bitkeks:
Done!

PR #177, opened 2018-08-21
Improve French translation

gnoxr:
Hi, thank you for this amazing project! :+1:

I believed I could help improving the French translation.
Detailed changes (also described in the commit message):

- Consistently use infinitive for action strings (Commentez → Commenter).
- Fix a few typos and simplify inclusive writing (aut⋅eur⋅ice⋅s → auteur⋅ice⋅s).
- End more error messages with a period to make forms more consistent.
- Translate "Previous page" and "Next page".

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -78,7 +78,7 @@ msgstr "Cet article est placé sous la licence {{ license }}"
 
 msgid "One like"
 msgid_plural "{{ count }} likes"
-msgstr[0] "Une personne aime cet article"
+msgstr[0] "{{ count }} personne aime cet article"
 msgstr[1] "{{ count }} personnes aiment cet article"
 
 msgid "I don&#x27;t like this anymore"
@@ -165,7 +165,7 @@ msgstr "Récemment repartagé"
 
 msgid "One follower"
 msgid_plural "{{ count }} followers"
-msgstr[0] "Un⋅e abonné⋅e"
+msgstr[0] "{{ count }} abonné⋅e"
 msgstr[1] "{{ count }} abonné⋅e⋅s"
 
 msgid "Edit your account"
@@ -326,7 +326,7 @@ msgstr "Le mot de passe doit faire au moins 8 caractères."
 
 msgid "One author in this blog: "
 msgid_plural "{{ count }} authors in this blog: "
-msgstr[0] "Un⋅e auteur⋅ice dans ce blog : "
+msgstr[0] "{{ count }} auteur⋅ice dans ce blog : "
 msgstr[1] "{{ count }} auteur⋅ice⋅s dans ce blog : "
 
 msgid "Login or use your Fediverse account to interact with this article"
@@ -340,7 +340,7 @@ msgstr "Optionnel"
 #, fuzzy
 msgid "One article in this blog"
 msgid_plural "{{ count }} articles in this blog"
-msgstr[0] "Un article dans ce blog"
+msgstr[0] "{{ count }} article dans ce blog"
 msgstr[1] "{{ count }} articles dans ce blog"
 
 msgid "Previous page"
--- a/po/fr.po
+++ b/po/fr.po
@@ -43,18 +43,18 @@ msgid "Submit comment"
 msgstr "Envoyer le commentaire"
 
 msgid "Something broke on our side."
-msgstr "Nous avons cassé quelque chose"
+msgstr "Nous avons cassé quelque chose."
 
 msgid "Sorry about that. If you think this is a bug, please report it."
 msgstr ""
 "Nous sommes désolé⋅e⋅s. Si vous pensez que c'est un bogue, merci de le "
-"rapporter."
+"signaler."
 
 msgid "Configuration"
 msgstr "Configuration"
 
 msgid "Configure your instance"
-msgstr "Configurez votre instance"
+msgstr "Configurer votre instance"
 
 msgid "Name"
 msgstr "Nom"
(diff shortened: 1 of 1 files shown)


po/fr.po
```diff
@@ -78,7 +78,7 @@ msgstr "Cet article est placé sous la licence {{ license }}"
 
 msgid "One like"
 msgid_plural "{{ count }} likes"
-msgstr[0] "{{ count }} personne aime cet article"
+msgstr[0] "Une personne aime cet article"
```
fdb-hiroshima: This fail when there are 0 likes, as in french 0 is singular it is treated by this rule. If a post is liked by nobody it will still show "Une personne aime cet article"

po/fr.po
```diff
@@ -319,15 +319,15 @@ msgid "Username can't be empty"
...
+msgstr "Adresse mail invalide."
 
 msgid "Password should be at least 8 characters long"
 msgstr "Le mot de passe doit faire au moins 8 caractères."
 
 msgid "One author in this blog: "
 msgid_plural "{{ count }} authors in this blog: "
-msgstr[0] "{{ count }} aut⋅eur⋅rice dans ce blog : "
-msgstr[1] "{{ count }} aut⋅eur⋅rice⋅s dans ce blog : "
+msgstr[0] "Un⋅e auteur⋅ice dans ce blog : "
```
fdb-hiroshima: Same here

po/fr.po
```diff
@@ -340,14 +340,14 @@ msgstr "Optionnel"
 #, fuzzy
 msgid "One article in this blog"
 msgid_plural "{{ count }} articles in this blog"
-msgstr[0] "{{ count }} aut⋅eur⋅rice dans ce blog : "
-msgstr[1] "{{ count }} aut⋅eur⋅rice⋅s dans ce blog : "
+msgstr[0] "Un article dans ce blog"
```
fdb-hiroshima: Same here again

po/fr.po
gnoxr: Thanks for your feedback. My bad, I thought `msgstr[0]` was for the singular `msgid` and `msgstr[1]` for `msgid_plural`. Is there a way to have a string for zero, a string for one, and a string for *n*?

French behaves a bit strangely; whereas the number 1 is singular, the number 0 is plural unless we replace it with an expression like “_Aucun⋅e_” (None) or “_Personne_” (No one) to make it singular.

> _0 personne**s** aime**nt** cet article_ or _Personne n'aime cet article_
> _Une personne aime cet article_
> _2(+) personne**s** aime**nt** cet article_

I would avoid “_Personne n'aime cet article_” since it translates to “No one likes this article” which could make authors feel quite bad, unless we use the past tense “_Personne n'a aimé cet article_” (“No one liked this article”), which suggests a “_pour l'instant_”/“yet”.

Would something like that work?

```
msgid "One like"
msgid_plural "{{ count }} likes"
msgstr[0] "Personne n'a aimé cet article"
msgstr[1] "Une personne aime cet article"
msgstr[2] "{{count}} personnes aiment cet article"
```

It would be somehow inconsistent though to use the past tense for the zero-case and the present tense for the others. Perhaps we could consider using the past tense everywhere?

po/fr.po
```diff
@@ -165,7 +165,7 @@ msgstr "Récemment repartagé"
 
 msgid "One follower"
 msgid_plural "{{ count }} followers"
-msgstr[0] "{{ count }} abonné⋅e"
+msgstr[0] "Un⋅e abonné⋅e"
```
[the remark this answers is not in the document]
gnoxr: Similar answer as above (zero is plural so `msgstr[0] "{{ count }} abonné⋅e"` won't work), what do you think of:

```
msgid "One follower"
msgid_plural "{{ count }} followers"
msgstr[0] "Pas d'abonné⋅e⋅s"
msgstr[1] "Un⋅e abonné⋅e"
msgstr[2] "{{count}} abonné⋅e⋅s"
```

We should consider using only numbers depending on the use case. On some other social media, profile statistics are displayed only with numerals (0, 1, 2, 784, 3.6k) in order for them to be consistent when stacked next to each other. The word form is only used for zero and one in longer sentences (notifications).

The alternative would be:

```
msgid "One follower"
msgid_plural "{{ count }} followers"
msgstr[0] "0 abonné⋅e⋅s"
msgstr[1] "1 abonné⋅e"
msgstr[2] "{{count}} abonné⋅e⋅s"
```

po/fr.po
fdb-hiroshima: Zero is not plural in french, there are some situation where we use zero with a plural ("zéro sucres ajoutés") but the general case is zero is singular, ex "[zéro faute d'orthographe]( https://fr.wiktionary.org/wiki/z%C3%A9ro#Adjectif_num%C3%A9ral)"

To choose what is plural and singular, the i18n engine look at the header `Plural-Forms` (line 13), for some plural it's simply `n != 1` (English, German...), some are more complex (take a look at pl.po if you want, in that language there are 2 plurals and one singular)

Even if I'm a native french speaker I've a hard time knowing what might be hurting peoples, I'm not exactly of the social kind. I guess you can do whatever seems best to translate without hurting people's feelings. (If you want to translate with more than 2 forms, i.e. zero, one and more, tell me and I will send you the corresponding `Plural-Forms`, but then you will have to update every msgstr[X] to the new plural, you did not update reshages translation previously)

gnoxr:
After looking up you're right, zero is singular. Being a French native didn't help me there! :p I'm reverting the changes for those lines. I believe there is nothing harsh in the current version (regarding my comment about avoiding “No one” + present tense).

PR #178, opened 2018-08-30
151 - Make the editor bigger and autosize it

### LINKED ISSUE #151: Bigger editor
Make it higher by default, and make it expand as you type more text.

Ilphrin:
Fixes #151 

I added bunch of JavaScript code, to make the `<textarea>` auto-resizeable. I don't know what are the code conventions here (is there one?)

### DIFF
--- a/static/js/autoExpand.js
+++ b/static/js/autoExpand.js
@@ -1,0 +1,14 @@
+function autosize(){
+  const el = this;
+  el.style.height = 'auto';
+  el.style.height = (el.scrollHeight ) + 'px';
+}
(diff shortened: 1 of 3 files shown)


Ilphrin:
I didn't know about this standard, if you want I can add a commit to make changes following this convention? ^^

BaptisteGelez:
It is not really important. I think I will explicitly write which standards we are using for each language, and then fix files that are not conforming. But if you have the will to change your code, feel free to open another PR. :slightly_smiling_face:

PR #205, opened 2018-09-07
Update to a more recent rocket and rust toolchain

lthms:
So, I have made some progress tonight, but unfortunately, this has become way more complicated than initially expected. The main reason is pretty simple: rocket has introduced some serious breaking changes related to the `Uri` type, and I am not familiar enough unfortunately to fix them easily.

I push this so you folks can have a look, maybe with enough familiarity this is trivial to fix.

Beware that this PR uses my forks of `rocket_{i18n,csrf}`. They are basically working, AFAICT at least, but I wanted to have a working Plume before creating PR there.

So, yeah. TL;DR: I tried and failed, but maybe what I achieved will help you?

To do before merging:
- [x] Fix the warnings
- [x] PR to `rocket_i18n`
- [x] PR to `rocket_csrf`


lthms:
I will try to do that, if that’s okay with you.

BaptisteGelez:
Yes, of course! If you have any question about the code, you know you can ask me :wink:

lthms:
I’ve pushed a new commit that fixes the remaining errors. There remains a lot of warnings, though.

To do before merging:
 - [ ] PR to rocket_i18n
- [ ] PR to rocket_csrf
- [ ] Fix the warnings

lthms:
Some warnings are due to a change in `rustc` that affects `diesel`. The crate has already been updated, but not released.

BaptisteGelez:
@lthms Could you please use the git repository directly in the Cargo.toml then, with a pinned `rev`? We will go back to a stable version once it will be released.

lthms:
I think the PR of interest is https://github.com/diesel-rs/diesel/pull/1787

So we need to depend on at least this revision. Unfortunately, I don't know if there are pending issues in current `diesel` `master` branch, so I have to admit I am not sure which revision to choose…

BaptisteGelez:
Another solution could be to ignore these warnings for the moment… https://github.com/rust-lang/crates.io/pull/1469/files

lthms:
Once https://github.com/fdb-hiroshima/rocket_csrf/pull/1 is merged, I will update `Cargo.toml` to use the upstream `rocket_csrf`, squash my commits and I think it will be okay, at least from my side.

lthms:
From my perspective, this PR is now ready. I don’t think I have broken something, but I guess you never know :D.

PR #212, opened 2018-09-10
Add webapp manifest

zcdunn:
Add a simple [Webapp Manifest](https://developer.mozilla.org/en-US/docs/Web/Manifest).

This enables users to add a shortcut to their homescreen/desktop.


zcdunn:
😓 oops. I've got some errors to fix.

Issue #220, opened 2018-09-13
Markdown Links are magled on repeated editing of /admin


igalic:
After entering the following markdown into

```markdown
Esotericsystems is my non-technical, sometimes even creative out-let.

for technical musings look at [Brainsware](https://brainsware.at/blog/) or [Sealas](https://sealas.at/blog/)
```

trying to extend it later, it looked like so:

```markdown
Esotericsystems is my non-technical, sometimes even creative out-let.

for technical musings look at [Brainsware](https:&#x2F;&#x2F;brainsware.at&#x2F;blog&#x2F;) or [Sealas](https:&#x2F;&#x2F;sealas.at&#x2F;blog&#x2F;)
```

BaptisteGelez:
I think it may be because `Instance.short_description` and `Instance.long_description` are `String`s and not `SafeString`s.

igalic:
I'll try that out locally, and submit a patch if it works

igalic:
```diff
diff --git a/plume-models/src/blogs.rs b/plume-models/src/blogs.rs
index a992f74..9c73a7d 100644
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -22,6 +22,7 @@ use plume_common::activity_pub::{
     inbox::WithInbox,
     sign
 };
+use safe_string::SafeString;
 use instance::*;
 use users::User;
 use schema::blogs;
@@ -142,8 +143,8 @@ impl Blog {
                     name: inst.clone(),
                     local: false,
                     // We don't really care about all the following for remote instances
-                    long_description: String::new(),
-                    short_description: String::new(),
+                    long_description: SafeString::new(&<String>::new()),
+                    short_description: SafeString::new(&<String>::new()),
                     default_license: String::new(),
                     open_registrations: true,
                     short_description_html: String::new(),
diff --git a/plume-models/src/instance.rs b/plume-models/src/instance.rs
index 52830aa..a250306 100644
--- a/plume-models/src/instance.rs
+++ b/plume-models/src/instance.rs
@@ -3,6 +3,7 @@ use diesel::{self, QueryDsl, RunQueryDsl, ExpressionMethods, PgConnection};
 use std::iter::Iterator;
 
 use plume_common::utils::md_to_html;
+use safe_string::SafeString;
 use ap_url;
 use users::User;
 use schema::{instances, users};
@@ -16,8 +17,8 @@ pub struct Instance {
     pub blocked: bool,
[...]

PR #223, opened 2018-09-14
make blog/instance description a SafeString

igalic:
long_description & short_description's documentation say they can be
Markdown, but they are String, not SafeString.

This led to escaped strings being printed in the editor
https://github.com/Plume-org/Plume/issues/220


igalic:
n.b.: This is a work-in-progress and does ***not*** solve the problem yet

pwoolcoc:
For the `FromFormValue` problem, you should be able to delegate the parsing to the `String::from_form_value` method. I haven't compiled this but it should go something like:


```
// safe_string.rs
use rocket::request::FromFormValue;
use rocket::http::RawStr;

impl<'v> FromFormValue<'v> for SafeString {
    type Error = &'v RawStr;

    fn from_form_value(form_value: &'v RawStr) -> Result<SafeString, &'v RawStr> {
        let val = String::from_form_value(form_value)?;
        SafeString {
          value: val,
        }
    }
}
```

plume-models/src/blogs.rs
```diff
@@ -142,8 +143,8 @@ impl Blog {
                     name: inst.clone(),
                     local: false,
                     // We don't really care about all the following for remote instances
-                    long_description: String::new(),
-                    short_description: String::new(),
+                    long_description: SafeString::new(&<String>::new()),
```
pwoolcoc: you can change `&<String>::new` here to just `""`, since `SafeString::new()` takes a `&str`. It will save an allocation, at least, especially since currently the allocation just gets dropped and another is made in the `ammonia::clean` function.

plume-models/src/users.rs
```diff
@@ -205,8 +205,8 @@ impl User {
                     public_domain: inst.clone(),
                     local: false,
                     // We don't really care about all the following for remote instances
-                    long_description: String::new(),
-                    short_description: String::new(),
+                    long_description: SafeString::new(&<String>::new()),
```
pwoolcoc: Same comment here re: `String::new`

src/setup.rs
```diff
@@ -152,8 +153,8 @@ fn quick_setup(conn: DbConn) {
         public_domain: domain,
         name: name,
         local: true,
-        long_description: String::new(),
-        short_description: String::new(),
+        long_description: SafeString::new(&<String>::new()),
```
pwoolcoc: and here as well

igalic:
@pwoolcoc thank you for your review!
i've pushed one commit which addresses your concerns wrt allocation, and one which adds your `imp FromFormValue for SafeString`… now… to make use of it ;)

(by which i mean, "now to find out how to make use of it")

BaptisteGelez:
@igalic Thanks for working on this. Rocket will use your `FromFormValue` directly if you replace the `String` fields of the form struct (`InstanceSettingsForm` in `src/routes/instance.rs`) by `SafeString`.

plume-models/src/safe_string.rs
```diff
@@ -101,3 +101,17 @@ impl AsRef<str> for SafeString {
...
+use rocket::http::RawStr;
+
+impl<'v> FromFormValue<'v> for SafeString {
+    type Error = &'v RawStr;
+
+    fn from_form_value(form_value: &'v RawStr) -> Result<SafeString, &'v RawStr> {
+        let val = String::from_form_value(form_value)?;
+        Ok(SafeString {
+          value: val,
+        })
```
fdb-hiroshima: `SafeString`s are cleaned at creation, you must either call `SafeString::new(&str)` or `ammonia::clean(&str)` (which is used internally by new) when creating such struct, not built it directly from an unstrusted String.

plume-models/src/safe_string.rs
```diff
@@ -101,3 +101,15 @@ impl AsRef<str> for SafeString {
         &self.value
     }
 }
+
+use rocket::request::FromFormValue;
+use rocket::http::RawStr;
+
+impl<'v> FromFormValue<'v> for SafeString {
+    type Error = &'v RawStr;
```
igalic: can someone help me understand what `'v` means?

plume-models/src/safe_string.rs
igalic: this should be fixed in d62c72d

plume-models/src/safe_string.rs
fdb-hiroshima: `'v` is a lifetime, because you got a reference to a `RawStr`, not a `RawStr` directly. This helps Rust figure out how long a struct containing pointers (references) can live, and when some parts of it might got dropped and it's therefore no longer safe to access a struct. You can read more about it in [rust's book](https://doc.rust-lang.org/1.9.0/book/lifetimes.html), which probably explain better than me :wink:

igalic: okay, thanks! i was just not clear on the notation

igalic:
okay! so with the last commit (https://github.com/Plume-org/Plume/pull/223/commits/06718a5c8a0af6c9facd10debfadae100fd25aa2) i've now done everything everyone on here has so helpfully suggested

which means that it now… still doesn't work 😅

fdb-hiroshima:
I think you did all that was needed on rust side. Now you just have to edit templates to add `| safe` whenever one of the vars you changed to SafeString is used as done here for user's summary https://github.com/Plume-org/Plume/blob/d355379e01006938ef87d230b456509a9731294c/templates/users/header.html.tera#L41
This should be in `templates/instance/admin.html.tera` I think.

igalic:
but that's not done with content:
https://github.com/Plume-org/Plume/blob/d355379e01006938ef87d230b456509a9731294c/templates/posts/new.html.tera#L31

fdb-hiroshima:
The exact line you need is `<textarea id="short_description" name="short_description">{{ form.short_description | default(value=instance.short_description | safe) }}</textarea>`, with the  `| safe` inside `default`'s brackets. The difference with post's content is `instance.short_description` is read by `default`, not passed directly (same goes with `long_description`). I'm sorry I was not clear enough in my previous comment

PR #226, opened 2018-09-15
Add SQLite as supported database

igalic:
Extend Plume, or rather, its use of Diesel to support SQLite


igalic:
n.b.: ***this is a work in progress***

plume-models/src/blogs.rs
```diff
@@ -66,32 +66,32 @@ impl Blog {
...
         Instance::get(conn, self.instance_id).expect("Couldn't find instance")
     }
 
-    pub fn list_authors(&self, conn: &PgConnection) -> Vec<User> {
+    pub fn list_authors(&self, conn: &Connection) -> Vec<User> {
         use schema::blog_authors;
         use schema::users;
         let authors_ids = blog_authors::table.filter(blog_authors::blog_id.eq(self.id)).select(blog_authors::author_id);
-        users::table.filter(users::id.eq(any(authors_ids)))
+        users::table.filter(users::id.eq_any(authors_ids))
```
igalic: from the diesel docs:
> On PostgreSQL, you should use eq(any()) instead. This method may change in the future to automatically perform = ANY on PostgreSQL


have you noticed any performance degradation so far?

plume-models/src/blogs.rs
```diff
@@ -185,43 +185,43 @@ impl Blog {
...
     }
 
-    pub fn update_boxes(&self, conn: &PgConnection) {
+    pub fn update_boxes(&self, conn: &Connection) {
         let instance = self.get_instance(conn);
         if self.outbox_url.len() == 0 {
             diesel::update(self)
                 .set(blogs::outbox_url.eq(instance.compute_box(BLOG_PREFIX, self.actor_id.clone(), "outbox")))
-                .get_result::<Blog>(conn).expect("Couldn't update outbox URL");
+                .execute(conn).expect("Couldn't update outbox URL");
```
igalic: what's the difference between these methods?

plume-models/src/lib.rs
```diff
@@ -63,22 +70,41 @@ macro_rules! get {
...
-                .expect(concat!("Error updating ", stringify!($table)))
+                .execute(conn)
+                .expect(concat!("Error updating ", stringify!($table)));
+            Self::get(conn, self.id)
+                .expect(concat!(stringify!($table), " we just updated doesn't exist anymore???"))
+        }
+    };
+}
+
+macro_rules! last {
```
igalic: how come none of these macros have docs?

plume-models/src/blogs.rs
BaptisteGelez: No, and actually, for now `eq_any()` is equivalent to `eq(any())` with Postgres, but also supports SQlite.

plume-models/src/blogs.rs
BaptisteGelez: `get_result`(`s`) requires the DB backend to support `RETURNING` statements (so, only with Postgres). Too support both SQlite and Postgres we should emulate this function by first using `execute`, then loading the result in another query.

plume-models/src/lib.rs
BaptisteGelez: I'll add them! :grin:

igalic:
So far so good!

`setup.rs` is still a little irritating

building it — correctly — feels irritating. i'd much rather `setup.rs` be able to deal with both postgres & sqlite built-in at the same time.

PR #238, opened 2018-09-19
Fix CSRF issues

BaptisteGelez:
GET routes are not protected against CSRF. This commit changes the needed URLs to
POST and replace simple links with forms.

### DIFF
--- a/templates/posts/details.html.tera
+++ b/templates/posts/details.html.tera
@@ -12,7 +12,7 @@
 {% block content %}
     <h1 class="article">{{ article.post.title }}</h1>
     <h2 class="article">{{ article.post.subtitle }}</h2>
-    <p class="article-info">
+    <div class="article-info">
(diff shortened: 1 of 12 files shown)


templates/posts/details.html.tera
```diff
@@ -12,27 +12,29 @@
...
         &mdash;
         <span class="date">{{ date | date(format="%B %e, %Y") }}</span>
         {% if is_author %}
             &mdash;
             <a href="{{ article.url}}edit">{{ "Edit" | _ }}</a>
             &mdash;
-            <a href="{{ article.url}}delete" onclick="return confirm('Are you sure you?')">{{ "Delete this article" | _ }}</a>
+            <form class="inline" method="post" action="{{ article.url}}delete">
+                <input onclick="return confirm('Are you sure you?')" type="submit" value="{{ 'Delete this article' | _ }}">
+            </form>
```
fdb-hiroshima: It seems like onclick is not triggered on input elements, the article get deleted instantly with no warnings

fdb-hiroshima: actually this what a bad configuration from my side which did not allow inline js, it works as intended

PR #245, opened 2018-09-25
Some API endpoints for articles

BaptisteGelez:
- Define the API with canapi
- Implement `canapi::Provider` for `Post` (only `get` and `list` actually)
- Use these functions to serve the API (`/api/v1/posts/<id>` and `/api/v1/posts`)
- Document the API with Swagger (couldn't test it, I hope it will be served correctly :confused:)

It is a first step for #5


plume-models/src/posts.rs
```diff
@@ -55,6 +57,54 @@ pub struct NewPost {
...
+        let mut query = posts::table.into_boxed();
+        if let Some(title) = filter.title {
+            query = query.filter(posts::title.eq(title));
+        }
+        if let Some(subtitle) = filter.subtitle {
+            query = query.filter(posts::subtitle.eq(subtitle));
+        }
+        if let Some(content) = filter.content {
+            query = query.filter(posts::content.eq(content));
+        }
```
igalic: what's the point of these queries and the override of the query variable?

igalic: aahh, you're fetching different parts of the same article, and stuffing them into different boxes of the query variable!

plume-api/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+    ($url:expr => $ep:ty) => {
+        impl Endpoint for $ep {
+            type Id = i32;
+
+            fn endpoint() -> &'static str {
+                $url
+            }
+        }
+    };
+}
```
igalic: sometimes i dream of lisp indentation style in Javascript. soon, I'll start dreaming of lisp indentation style in rust

po/gl.po
```diff
@@ -325,7 +325,8 @@ msgstr[1] "{{ count }} autoras en este blog: "
 
 msgid "Login or use your Fediverse account to interact with this article"
 msgstr ""
-"Conéctese ou utilice a súa conta no fediverso para interactuar con este artigo"
+"Conéctese ou utilice a súa conta no fediverso para interactuar con este "
+"artigo"
```
igalic: why does this file keep changing with every new build?

src/api/posts.rs
```diff
@@ -0,0 +1,21 @@
...
+use plume_models::db_conn::DbConn;
+use plume_models::posts::Post;
+
+#[get("/posts/<id>")]
+fn get(id: i32, conn: DbConn) -> Json<serde_json::Value> {
+    let post = <Post as Provider<PgConnection>>::get(&*conn, id).ok();
+    Json(json!(post))
+}
+
+// TODO: handle query params
```
igalic: such as?

docs/api.yaml
```diff
@@ -0,0 +1,45 @@
+openapi: "3.0"
```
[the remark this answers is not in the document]
BaptisteGelez: not yet, but it may be possible thanks to canapi…

po/gl.po
BaptisteGelez: it is updated every time you run Plume. However, extracting i18n related task to an external tool may be better… :thinking:

src/api/posts.rs
BaptisteGelez: For instance, if you `GET /api/v1/posts?title=Hello%20world`, only posts with the `Hello world` title should be shown (`title` is not the best example, but `author_id` could be interesting for instance)

plume-models/src/posts.rs
BaptisteGelez: actually, the `PostEndpoint` type is here to represent both the client's query and the result sent by the API. `filter` is what is requested by the client, so if any filter have been specified I update the SQL `query` to only return the requested posts.

BaptisteGelez: not sure if I'm clear, tell me if not :grin:

igalic:
i just realized how easy CLI mappings would be to implement, once the API is in place.

src/api/posts.rs
fdb-hiroshima: So is this (at least partially) blocked by #149, or not related?

BaptisteGelez: No, actually we just need to get the raw query params, and deserialize them to `PostEndpoint` with something like `serde_qs`.

PR #248, opened 2018-09-29
Bad commands in the update guide

maximesrd:


maximesrd:
Yes, but I don't know how I can make this more presentable.

maximesrd:
Ok, I remove them !

PR #256, opened 2018-10-03
Verify activity's signature

### LINKED ISSUE #10: Verify incoming activities signature

fdb-hiroshima:
Verify signature for incoming activity. This is still work in progress
Fix #10

Todo:
- [X] Verify HTTP signature
- [x] Verify JSON-ld signature

Federation status at this time:
- [X] itself
- [x] older Plume
- [x] other ActivityPub implementations (Mastoodn, Pleroma tested)

Side-note:
As this as the ability to totally break federation, we should wait to be sure it does work with at least Mastodon, Pleroma and older Plume before merging this

### DIFF
--- a/plume-common/src/activity_pub/mod.rs
+++ b/plume-common/src/activity_pub/mod.rs
@@ -104,11 +104,12 @@ pub fn broadcast<S: sign::Signer, A: Activity, T: inbox::WithInbox + Actor>(send
 
     for inbox in boxes {
         // TODO: run it in Sidekiq or something like that
+        let mut headers = request::headers();
+        headers.set(request::digest(signed.to_string()));
         let res = Client::new()
             .post(&inbox[..])
-            .headers(request::headers())
-            .header(request::signature(sender, request::headers()))
-            .header(request::digest(signed.to_string()))
+            .headers(headers.clone())
+            .header(request::signature(sender, headers))
             .body(signed.to_string())
             .send();
         match res {
--- a/plume-common/src/activity_pub/mod.rs
+++ b/plume-common/src/activity_pub/mod.rs
@@ -1,10 +1,8 @@
 use activitypub::{Activity, Actor, Object, Link};
 use array_tool::vec::Uniq;
-use base64;
 use reqwest::Client;
 use rocket::{
-    Outcome,
-    http::{Status,HeaderMap},
+    Outcome, http::Status,
     response::{Response, Responder},
     request::{FromRequest, Request}
 };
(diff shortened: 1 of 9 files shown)


src/routes/instance.rs
```diff
@@ -189,13 +191,41 @@ fn ban(_admin: Admin, conn: DbConn, id: i32) -> Redirect {
     Redirect::to(uri!(admin_users))
 }
 
+struct Headers<'r> {
+    headers: HeaderMap<'r>,
+}
```
fdb-hiroshima: This should probably go somewhere else, as it will be used on both the shared inbox and the personal one, which is defined in another file

src/routes/instance.rs
```diff
@@ -189,13 +191,41 @@ fn ban(_admin: Admin, conn: DbConn, id: i32) -> Redirect {
...
-fn shared_inbox(conn: DbConn, data: String) -> String {
+fn shared_inbox(conn: DbConn, data: String, headers: Headers) -> String {
     let act: serde_json::Value = serde_json::from_str(&data[..]).unwrap();
 
     let activity = act.clone();
     let actor_id = activity["actor"].as_str()
         .unwrap_or_else(|| activity["actor"]["id"].as_str().expect("No actor ID for incoming activity, blocks by panicking"));
+
+    let sig = match verify_http_headers(&User::from_url(&conn, actor_id.to_owned()).unwrap(), headers.headers, data) {
+        SignatureValidity::Valid => true,
```
fdb-hiroshima: The same as to be added on personal inboxes

plume-common/src/activity_pub/mod.rs
```diff
@@ -128,6 +128,15 @@ pub enum SignatureValidity {
     Absent,
 }
 
+impl SignatureValidity {
+    pub fn is_secure(&self) -> bool {
+        match self {
+            SignatureValidity::Valid => true,
+            _ => false,
```
BaptisteGelez: Why not `self == SignatureValidity::Valid`?

fdb-hiroshima: Because `PartialEq` is not implemented for SignatureValidity? Joking, it can be derived. I just copied the whole match from [src/instance.rs](https://github.com/Plume-org/Plume/blob/0a5d435249f9f4b4df7f42fc40578af302a91954/src/routes/instance.rs#L218), moved the todo somewhere-else, and basically forgot there were an shorter and cleaner way

plume-common/src/activity_pub/mod.rs
```diff
@@ -118,6 +120,68 @@ pub fn broadcast<S: sign::Signer, A: Activity, T: inbox::WithInbox + Actor>(send
...
+                }
+            } else {
+                SignatureValidity::Invalid
+            }
+        } else {
+            SignatureValidity::Invalid
+        }
+    } else {
+        SignatureValidity::Absent
+    }
```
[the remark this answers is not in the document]
fdb-hiroshima: I don't know if it's frowned upon, but this is something that I don't like personally, however I agree this would be much easier to read with early returns, I'll modify that at some point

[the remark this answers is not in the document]
igalic: thank you for explaining
maybe we can find a better way to deal with the trees ;)
maybe it's just me (or my dyslexia), but i find them very hard to read and follow

igalic:
i'm kinda surprised no one's written a json-ld crate so far

plume-common/src/activity_pub/mod.rs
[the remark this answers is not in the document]
fdb-hiroshima: @igalic I tried to flatten the tree as you suggested, you might want to take a look and tell me if it's better

fdb-hiroshima:
@BaptisteGelez is it possible you try to deploy this branch on your test instance to test federation with Mastodon, Pleroma and maybe other activity-pub enabled soft?

BaptisteGelez:
@fdb-hiroshima baptiste.gelez.xyz is now running this branch. If you need anything else (logs for instance) tell me.

fdb-hiroshima:
@BaptisteGelez could you update your instance to the latest commit, and if it still don't work, give me back the generated logs?

Issue #157, opened 2018-07-21
Authentication failed for user "plume"


Ilphrin:
Hi,

After running `cargo run`, Plume create a 'plume' user and a database associated to it. But after setting the password, Plume says that it can't authenticate, as if the password was wrong (For test purpose, I tried with a simple password '**coucou**' without any special characters):

```
Finished dev [unoptimized + debuginfo] target(s) in 0.17s
     Running `target/debug/plume`



Welcome in the Plume setup tool.
It will help you setup your new instance, by asking you a few questions.
Then you'll be able to enjoy Plume!

First let's check that you have all the required dependencies. Press Enter to start.

  ✔️ All native dependencies are present.

We are going to setup the database.

About to create a new PostgreSQL user named 'plume'

Enter password for new role: 
Enter it again: 
  ✔️ Done

About to create a new PostgreSQL database named 'plume'

  ✔️ Done

Running migrations…

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ConnectionError(BadConnection("FATAL:  password authentication failed for user \"plume\"\nFATAL:  password authentication failed for user \"plume\"\n"))', libcore/result.rs:945:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
thread 'main' panicked at 'Couldn't init DB pool', libcore/option.rs:960:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

OS: Pop_OS! 18.04 (Ubuntu) 64 bits

Running with latest nightly rustc (with rustup) version

Ilphrin:
Oh I see ^^ If I have some time I'll try to take a look at this (after installing a trying plume x) )

jeroenpraat:
I have the same problem. 

I think it has to do with the fact that the system user 'plume' has not the permission to use Postgresql's createuser and createdb.

I also tried to create the user and database as the system user 'postgres', but when I run 'cargo run' again it still complains like this:

```
About to create a new PostgreSQL user named 'plume'

Enter password for new role: 
Enter it again: 
createuser: creation of new role failed: ERROR:  permission denied to create role
About to create a new PostgreSQL database named 'plume'

createdb: database creation failed: ERROR:  database "plume" already exists
```
So I'm stuck now at installing Plume. Please help?!

BaptisteGelez:
@jeroenpraat A workaround would be to start Plume with this command:

```bash
DB_URL=postgres://USER:PASSWORD@localhost/plume cargo run
```

[...]

Issue #209, opened 2018-09-08
Update without script


anmol26s:
Hi
I would like to upgrade without git as mentioned here https://github.com/Plume-org/Plume/blob/master/docs/UPDATE.md.

So I would like to : 
1. Backup the important files and folders like .env, etc.
2. Delete the the source folder
3. Download and decompress the fresh sources to the folder.
4. Restore the importation files and folders like .env to the new unpacked sources.

So for this I would like to know what files and folders are needed to be backed up and needs to be restored to the newly unpacked sources, so that the Plume don't break.

BaptisteGelez:
The only configuration file is the `.env`, everything else is stored in the DB, or will be generated by Plume at runtime (translations). If you write a Bash script to make this update automatic, please share it there, it could be useful to other persons (but there is no problem if you don't).

BaptisteGelez:
Wow, great! For the systemd issue, if I understand correctly `__FINAL_PATH__` is the path where Plume will be installed, right? Then, replacing

```
ExecStart=__FINALPATH__/ cargo run >> /var/log/__APP__/__APP__.log 2>&1
```

by

```
ExecStart=cargo run >> /var/log/__APP__/__APP__.log 2>&1
```

May work, I think, since there is no reason this path appears here. But I may be wrong. :man_shrugging:

anmol26s:
>  if I understand correctly __FINAL_PATH__ is the path where Plume will be installed, right?

yes you are right the __FINAL_PATH__ is the directory where Plume is installed.
I have already tried ExecStart=cargo run, but its not working? Any other advise?

anmol26s:
ExecStart=cargo run gives me error at the end of the complie:

`error: linking with `cc` failed: exit code: 1`

BaptisteGelez:
Do you have any other error message before that?

BaptisteGelez:
We recently updated the Rust Nightly we are using, you should re-try after updating to the version specified in the `rust-toolchain` file.

anmol26s:
Ok thanks
The build work now. But the diesel building by this command:
`cargo install diesel_cli --no-default-features --features postgres --version '=1.2.0'`

Gives a warning : 
`warning: be sure to add /root/.cargo/bin to your PATH to be able to run the installed binaries`

and migration is not working because it says  **diesel: command not found**

BaptisteGelez:
Is `/root/.cargo/bin` in the `PATH`? I guess the warning is shown because it is not.

[...]

Issue #89, opened 2018-06-27
Setup: should allow to set DB host and port ( remote DB / pgbouncers / ... )


gled-rs:
I may be wrong but setup currently tries to setup pgsql on the local server.

If pg is on a different box, or you need to use pgbouncer on the non standard port, it fails.

We could have the option of creating the user by hand and set the DB_URL ourselves.

gled-rs:
For now if someone needs to do it:
- create your user and db out of plume
- patch src/setup.rs:

pub fn check() -> PgPool {
    if let Some(pool) = init_pool() {
        match pool.get() {
            Ok(conn) => {
                let db_conn = DbConn(conn);
//                if Instance::get_local(&*db_conn).is_none() {
                    migrate();
                    run_setup(Some(db_conn));
//                }
            }
- DB_URL=postgres://user:pass@server/db cargo run
- remove the patch on src/setup.rs and rerun.

Now to get rid of the po files not found even if they are here :)

gled-rs:
workaround in #99 :)

PR #261, opened 2018-10-04
CLI tools

### LINKED ISSUE #260: Improve instance setup
The current setup script is not working on many setups, makes it hard to create packages for Yunohost (for instance), and does more harm than good: it should be replaces with something else. The decision that was taken on Loomio was to have non-interactive tools with an interactive overlay. How I see this feature in details is:

- have a CLI tool to do a specific task, that accepts some arguments
- if one of these arguments is required but not specified, ask the user to input it

The tools that would be needed to replace the current setup script (we could add more of them after):
- [ ] `plm instance new`
- [ ] `plm user new`

[Loomio debate](https://framavox.org/d/1S4Itx6R/the-setup-script)

BaptisteGelez:
- New sub-crate: plume-cli
- Creates a binary called 'plm'
- Uses clap to parse CLI arguments
- Add a first command: plm instance new (to init the local instance)

Example usage:

```
plm instance new -d foo.bar -n 'Hey'
```

Fix #260


plume-cli/src/instance.rs
```diff
@@ -0,0 +1,63 @@
+use clap::{Arg, ArgMatches, App, SubCommand};
+use diesel::PgConnection;
```
igalic: shouldn't this be database agnostic now?

BaptisteGelez: Oops, forgot it! :grin:

BaptisteGelez: Fixed

BaptisteGelez:
Will also fix #157 #89 and #209

docs/CLI.md
```diff
@@ -0,0 +1,51 @@
...
+
+```bash
+plm instance new --private --domain plu.me --name 'My Plume Instance' -l 'CC-BY'
+```
+
+**Arguments:**
+
+- `--domain`, `-d`: the domain name on which your instance will be available.
+- `--name`, `-n`: The name of your instance. It will be displayed on the homepage.
+- `--default-license`, `-l`: the license to use for articles written on this instance, if no other license is explicitely specified. Optional, defaults to CC-0.
```
igalic: this is not correct anymore

docs/INSTALL.md
```diff
@@ -133,44 +133,73 @@ This command may be useful if you decided to use a separate database server.
 
 ## Starting Plume
```
igalic: We should have fixed the info on how to compile Plume with sqlite in #226, why DM i only thinking of this now?

plume-cli/src/instance.rs
```diff
@@ -0,0 +1,63 @@
...
+        _ => println!("Unknwon subcommand"),
+    }
+}
+
+fn new<'a>(args: &ArgMatches<'a>, conn: &Connection) {
+    let domain = args.value_of("domain").map(String::from)
+        .unwrap_or_else(|| env::var("BASE_URL")
+            .unwrap_or_else(|_| super::ask_for("Domain name")));
+    let name = args.value_of("name").map(String::from).unwrap_or_else(|| super::ask_for("Instance name"));
+    let license = args.value_of("default-license").map(String::from).unwrap_or(String::from("CC-0"));
```
igalic: we've decided to change the default license

igalic:
we have forgotten to fix up the compile / installation instructions after/with #226

BaptisteGelez:
@igalic yes, I thought about it a few minutes ago too. I will update them right now.

docs/CLI.md
BaptisteGelez: Yes, but I will change everything in the same PR, so for now I used CC-0.

PR #264, opened 2018-10-06
Missing dependencies

maximesrd:
sqlite was missing, which caused the compilation to fail, even if the database used is postgresql


BaptisteGelez:
Ahah, I just made #265 which adds it too, but it looks like I forgot some of them. I'll merge our two branches. Thank you.

PR #271, opened 2018-10-08
Update French translation

yannicka:


yannicka:
Pas de quoi. :smiley:

PR #281, opened 2018-10-20
Panic less and with better messages

### LINKED ISSUE #21: Avoid panics
I think we are `unwrap`ing too often, when we could show actual error pages, with a description of the message.

fdb-hiroshima:
Fix #21 
Change most unwrap to expect, and don't panic when it can be avoided. Also return better http codes, i.e. 400 and 404, when appropriate.
The trait `rocker::request::FromParam` should probably be implemented for Post, Blog, Media, User and Tag. Then we wouldn't need to return `Option`s whenever an entity does not exist, and just let Rocket do the work for us

### DIFF
--- a/plume-cli/src/main.rs
+++ b/plume-cli/src/main.rs
@@ -27,7 +27,7 @@ fn main() {
     match matches.subcommand() {
         ("instance", Some(args)) => instance::run(args, &conn.expect("Couldn't connect to the database.")),
         ("users", Some(args)) => users::run(args, &conn.expect("Couldn't connect to the database.")),
-        _ => app.print_help().unwrap()
+        _ => app.print_help().expect("Couldn't print help")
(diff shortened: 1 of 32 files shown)


BaptisteGelez:
And yes, implementing FromParams for these types would be a good idea.

fdb-hiroshima:
After trying to, I guess it might not be possible, because we need a db connection, and I don't think there is a way to access managed state from from_param (it only takes a &RawStr as parameters). Nonetheless I'll ask on Matrix

fdb-hiroshima:
If we end up doing manual routing it may reduce code duplication a bit, but it will add a lot of complexity, and we won't be able to use rocket's code generation if we do so. I'll see what answer I get from Matrix, but what I said earlier might very well be a false good idea

PR #282, opened 2018-10-20
Add support for blog deletion

### LINKED ISSUE #181: Blog deletion

fdb-hiroshima:
fix #181

At the moment it is done in a similar way to user deletion, i.e. post deletion is not federated, whereas it should probably be. Tell me if I have to implement that part too before merging.

There is a new string to translate ("Delete this blog") however I don't know well how gettext work so I did not add it to `.po`s

### DIFF
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -17,13 +17,14 @@ use webfinger::*;
 use {BASE_URL, USE_HTTPS, Connection};
 use plume_common::activity_pub::{
     ap_accept_header, ApSignature, ActivityStream, Id, IntoId, PublicKey,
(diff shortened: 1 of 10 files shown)


BaptisteGelez:
I think article deletion should be federated, but we could probably implement it in another PR.

To add the new string to the `.po` files, just add this to `po/plume.pot` (the template used for the `.po`)

```po
msgid "Delete this blog"
msgstr ""
```

fdb-hiroshima:
I was testing if changes I was doing would work, then I noticed I did not have entry from previous attempts. Delete is set to cascade on blog_authors, so we don't actually need to delete blog_authors, databases do it by themselves

Issue #170, opened 2018-08-03
Make login failure time consistent


fdb-hiroshima:
Login failure due to user not being found is pretty much instantaneous, where error due to invalid password take a quite long time. When we detect this is not a known user, we should compute the same password derivation function on the provided password to prevent disclosure of informations (and not really related but the cost used for password derivation should be cheaper, 6 full seconds (measured on the demo instance) of cpu usage is quite expensive)

PR #285, opened 2018-10-21
API Authentication

### LINKED ISSUE #275: API authentication
Currently the API doesn't require authentication. It is not a real problem yet, as the two routes we implemented are read-only, but for routes that will add/change user content, request should be authenticated.

Here is how I see it (tell me if you think something is wrong with this design, I'm not used REST APIs):

- There is an endpoint in the API for apps to register themselves, it returns a client ID, a client secret
- there is another endpoint to get an access token, that requires the client ID, secret and users credentials
- this token should be sent in the `Authorization` header for authenticated API requests
- Users can revoke any token from Plume

BaptisteGelez:
- [x] `App` model
- [x] API endpoint to register an app
- [x] `ApiToken` model
- [x] OAuth2 endpoint to get a token, from app and user credentials
- [x] make `ApiToken` a request guard

Fixes #275


src/main.rs
```diff
@@ -176,6 +178,7 @@ fn main() {
                     ("/@/<name>/inbox".to_owned(), "/@/<name>/inbox".to_owned(), rocket::http::Method::Post),
                     ("/login".to_owned(), "/login".to_owned(), rocket::http::Method::Post),
                     ("/users/new".to_owned(), "/users/new".to_owned(), rocket::http::Method::Post),
+                    ("/api/v1/<endpoint>".to_owned(), "/api/v1/<endpoint>".to_owned(), rocket::http::Method::Post)
```
fdb-hiroshima: I think actually the whole `/api` can be authorized, if I remember well it's denoted by `/api/<path..>`

plume-models/src/apps.rs
```diff
@@ -0,0 +1,80 @@
+use canapi::{Error, Provider};
```
igalic: should i be watching the canapi repo as well?

BaptisteGelez: As you want

docs/API.md
```diff
@@ -1,3 +1,36 @@
+# API documentation
+
+## Getting an API token
+
+To get access to the API, you should register your app and obtain a
+token. To do so, use the `apps` API (accessible without a token) to create
```
fdb-hiroshima: Just "use apps" is maybe a bit unclear, maybe you should use the full path, and also tell how to use it (Do a post with such data, and such other is optional and....)
```suggestion
token. To do so, use the `/api/v1/apps` API (accessible without a token) to create
```

plume-api/src/apps.rs
```diff
@@ -0,0 +1,13 @@
+use canapi::Endpoint;
+
+#[derive(Clone, Default, Serialize, Deserialize)]
+pub struct AppEndpoint {
```
fdb-hiroshima: This feel strange to be at the same time data received from Post (with id, client_id and client_secret ignored, as they must be generated by the server) and data returned by the api (with those same field used, and most likely different than what was originally posted if they where). It should either be 2 different struct or at least a struct with FromForm custom-implemented to ensure that

plume-api/src/apps.rs
```diff
@@ -0,0 +1,13 @@
...
+    pub id: Option<i32>,
+    pub name: Option<String>,
```
fdb-hiroshima: name seems to be required by https://github.com/Plume-org/Plume/blob/36297101f261f473fd03d86a22f46379d125e002/plume-models/src/apps.rs#L47
So I think it should not be an Option

src/api/mod.rs
```diff
@@ -1 +1,54 @@
...
+                    "token": token.value
+                }))
+            } else {
+                Json(json!({
+                    "error": "Wrong password"
+                }))
+            }
+        } else {
+            Json(json!({
+                "error": "Unknown user"
```
fdb-hiroshima: both error message ("Wrong password" and "Unknown user") should probably be merged, for similar reasons as #170

docs/API.md
BaptisteGelez: The request should be documented with Swagger (but it is broken for the moment :cry:)

plume-api/src/apps.rs
BaptisteGelez: We can't use `FromForm` in plume-api, or we would loose all the benefits of canapi.

But I think I may add a `Server`/`Client`/`Both` wrapper type to specify when a field is required and make it easier to check if something has been forgotten.

Or maybe canapi is just a bad idea and we should drop it... :thinking:

src/api/authorization.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn to_str() -> &'static str {
+        "posts"
+    }
+}
+
+// We have to use A and S in the struct definition
+// otherwise rustc complains they are useless
+//
+// A nicer solution is probably possible.
+pub struct Authorization<A, S> (Option<A>, Option<S>);
```
BaptisteGelez: I don't know if there is a better way to define this type. If I don't use `A` and `S` in its definition, it refuses to build.

src/api/authorization.rs
```diff
@@ -0,0 +1,57 @@
...
+where A: Action,
+      S: Scope
+{
+    type Error = ();
+
+    fn from_request(request: &'a Request<'r>) -> request::Outcome<Authorization<A, S>, ()> {
+        request.guard::<ApiToken>()
+            .map_failure(|_| (Status::Unauthorized, ()))
+            .and_then(|token| if token.can(A::to_str(), S::to_str()) {
+                Outcome::Success(Authorization(None, None))
```
BaptisteGelez: As you can see, these two Options will actually always be None, never Some(A) or Some(S).

src/api/authorization.rs
fdb-hiroshima: I don't think so

fdb-hiroshima: You can use PhantomData https://doc.rust-lang.org/beta/std/marker/struct.PhantomData.html, it'll probably do the trick
Simple example : https://gist.github.com/rust-play/9e51f5b8bb3a915a99d958f5ea982f1a

plume-models/src/api_tokens.rs
```diff
@@ -0,0 +1,88 @@
...
+            if s == what || s == full_scope {
+                return true
+            }
+        }
+        false
+    }
+
+    pub fn can_read(&self, what: &'static str) -> bool {
+        self.can("read", what)
+    }
```
fdb-hiroshima: This is confusing because `can` take a `what` set to "read" and a `scope` set to `what`. `what` should probably renamed `scope`, or something else should be renamed in `can`

plume-models/src/api_tokens.rs
```diff
@@ -0,0 +1,88 @@
...
+
+    pub fn can_write(&self, what: &'static str) -> bool {
+        self.can("write", what)
+    }
```
fdb-hiroshima: Same goes here

src/api/posts.rs
```diff
@@ -10,15 +10,16 @@ use plume_models::{
...
 #[get("/posts/<id>")]
-fn get(id: i32, conn: DbConn) -> Json<serde_json::Value> {
+fn get(id: i32, conn: DbConn, _auth: Authorization<Read, Post>) -> Json<serde_json::Value> {
     let post = <Post as Provider<Connection>>::get(&*conn, id).ok();
     Json(json!(post))
 }
 
 #[get("/posts")]
-fn list(conn: DbConn, uri: &Origin) -> Json<serde_json::Value> {
+fn list(conn: DbConn, uri: &Origin, _auth: Authorization<Read, Post>) -> Json<serde_json::Value> {
```
fdb-hiroshima: Same here

src/api/posts.rs
```diff
@@ -10,15 +10,16 @@ use plume_models::{
     db_conn::DbConn,
     posts::Post,
 };
+use api::authorization::*;
 
 #[get("/posts/<id>")]
-fn get(id: i32, conn: DbConn) -> Json<serde_json::Value> {
+fn get(id: i32, conn: DbConn, _auth: Authorization<Read, Post>) -> Json<serde_json::Value> {
```
fdb-hiroshima: this kind of endpoint can probably be called without tokens, at least as long as the post is published (and require a valid authorization, from a user having access to the post if it's not)

PR #293, opened 2018-10-25
Fix editor auto-expansion

### LINKED ISSUE #237: Edit mode : no side scroll yet / adaptative window along the composing text ?
Hi,
When in Edit mode, browsing our article is possible through the arrow keys only. 

Also, could the inner editing window grow automatically along the article length ?

Thanks

BaptisteGelez:
Fixes #237 

Another element with the `content` ID was present in the header.

### DIFF
--- a/templates/posts/new.html.tera
+++ b/templates/posts/new.html.tera
@@ -27,7 +27,7 @@
         {% endfor %}
     {% endif %}
 
-    <label for="content">{{ "Content" | _ }}<small>{{ "Markdown is supported" | _ }}</small></label>
+     <label for="plume-editor">{{ "Content" | _ }}<small>{{ "Markdown is supported" | _ }}</small></label>
     <textarea id="plume-editor" name="content" rows="20">{{ form.content | default(value="") }}</textarea>
 
     {{ macros::input(name="tags", label="Tags, separated by commas", errors=errors, form=form, optional=true) }}
--- a/templates/posts/new.html.tera
+++ b/templates/posts/new.html.tera
@@ -28,7 +28,7 @@
     {% endif %}
 
     <label for="content">{{ "Content" | _ }}<small>{{ "Markdown is supported" | _ }}</small></label>
-    <textarea id="content" name="content" rows="20">{{ form.content | default(value="") }}</textarea>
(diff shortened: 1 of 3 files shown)


fdb-hiroshima:
I've tried with a very big text (a few thousands words of lorem ipsum), the expansion works well, however I'm getting a weird effect when trying to edit anything behind the base size, it auto-scroll to the very beginning of the post at each key up, and go back to the cursor at each key down. The issue is reproducible on both Firefox (63) and Chromium (70)

BaptisteGelez:
I can't reproduce (I tried with Firefox 63 and Chromium 70 too :woman_shrugging:)

templates/posts/new.html.tera
```diff
@@ -28,7 +28,7 @@
     {% endif %}
 
     <label for="content">{{ "Content" | _ }}<small>{{ "Markdown is supported" | _ }}</small></label>
```
fdb-hiroshima: the `for` should be modified to match textarea's new id
```suggestion
     <label for="plume-editor">{{ "Content" | _ }}<small>{{ "Markdown is supported" | _ }}</small></label>
```

fdb-hiroshima:
@BaptisteGelez that's odd, I tried again on a fresh Plume install and this is definitely reproducible on my side. I tried one single big paragraph, multiple small ones, one word per line... anything longer than 50 lines seems to trigger it (which is in fact more than the hard-coded 20 rows)

BaptisteGelez:
OK, I can indeed reproduce with a longer text. I'll try to fix it.

BaptisteGelez:
The issue has normally been fixed (all these commits could probably be squashed before merging BTW)

fdb-hiroshima:
So it no longer scroll back and forth, however it do scroll on the first keypress after moving the cursor. This is way better and can probably be merged as is, but we might want to take a look at it. What I get is not coherent between browser however : on Firefox the line being edited is aligned with the bottom of the screen, whereas in Chromium it's aligned with the middle of it

BaptisteGelez:
I don't want to fight against browsers and their different implementation, so I will merge this branch as it is. If your issue is too problematic, we will still be able to fix it afterwards. This editor will (almost) disappear soon anyway…

PR #299, opened 2018-10-31
Article cover

### LINKED ISSUE #236: Picturing our Main titles
Looking to a Plume portal, none of the presented articles are appealling, they're visually very similar. A known distinguishment is to have a picture do its visual effect.

But we yet can't insert a picture in the title section.
Ideally, such a feature implementation would even possibly allow to mask the underlying declared textual title (this latter just helping Plume to sort things out) 

Plume v.0.2.0  plume.mastodon.host

BaptisteGelez:
Screenshots:

On the article page

On the home page

Fixes #236


BaptisteGelez:
I'm not 100% satisfied with the cover selector (a `<select>` field listing your media) but if we want to have a JavaScript-free implementation and to avoid duplicating the media upload form, I don't see any better solution. :woman_shrugging:

po/fr.po
```diff
@@ -630,3 +630,10 @@ msgstr "Cet article n’est pas encore publié."
 
 msgid "There is currently no article with that tag"
 msgstr "Il n'y a pas encore d'article avec ce tag"
+
+#, fuzzy
+msgid "Illustration"
+msgstr "Administration"
```
fdb-hiroshima: wow fuzzy translation is so wrong

BaptisteGelez:
Yes, I agree the way media are currently working is not ideal. I think we can change it once will we will switch to a new editor, by adding the possibility to upload an avatar directly from the profile edition page, and to add any other media directly from the editor.

fdb-hiroshima:
Anyway changing how we handle medias is neither the goal nor in the scope of this pr, so even if it's less than ideal, it shouldn't affect how we merge this pr (and actually making sure people are not using a song as a picture is not that important as in the end how they select those will be totally changed latter on)

templates/posts/new.html.tera
```diff
@@ -27,14 +27,24 @@
...
     {{ macros::input(name="tags", label="Tags, separated by commas", errors=errors, form=form, optional=true) }}
 
     {% set license_infos = "Default license will be {{ instance.default_license }}" | _(instance=instance) %}
     {{ macros::input(name="license", label="License", errors=errors, form=form, optional=true, details=license_infos) }}
 
+    <label for="cover">{{ "Illustration" | _ }}<small>{{ "Optional" | _ }}</small></label>
+    <select id="cover" name="cover">
+        <option value="none" {% if not form.cover %}selected{% endif %}>{{ "None" | _ }}</option>
+        {% for media in medias %}
+            <option value="{{ media.id }}" {% if form.cover == media.id %}selected{% endif %}>
```
fdb-hiroshima: This is the exact line that is failing. I think you can't compare `form.cover` because `form` is `null` when creating a new article. `if not form.cover` (2 lines above) is legal as it verify the existence of the field, and its "nullness", but a compare only make sense if the field exist, it being null or not

src/routes/posts.rs
```diff
@@ -206,9 +212,13 @@ fn update(blog: String, slug: String, user: User, conn: DbConn, data: LenientFor
             post.content = SafeString::new(&content);
             post.source = form.content.clone();
             post.license = license;
+            println!("Cover id from update: {:?}", form.cover);
+            post.cover_id = form.cover;
             post.update(&*conn);
+            println!("Cover id after update: {:?}", post.cover_id);
             let post = post.update_ap_url(&*conn);
 
+            println!("Cover id after after update: {:?}", post.cover_id);
```
fdb-hiroshima: Those 3 `println!` were for debug only I guess, so they should be removed

BaptisteGelez: Oops, you're right. :grimacing:

PR #306, opened 2018-11-05
i18n: Add Russian translation

link2xt:


link2xt:
@BaptisteGelez 
> I just saw that some messages are left untranslated, is there a reason for that?

Some entries are hard to translate, mostly because there is no common translation for some Fediverse terms yet. I have translated the rest but marked some entries as "needs work".

BaptisteGelez:
Okay, I will merge this as it is then. Thanks again!

PR #307, opened 2018-11-06
Post creation API

BaptisteGelez:


BaptisteGelez:
Ah, yes, I should write tests…

BaptisteGelez:
I think I will wait for #310 to be merged before adding tests, as it would probably create conflicts otherwise…

src/api/posts.rs
```diff
@@ -12,15 +13,144 @@ use plume_models::{
...
+            scopes: "write".to_string(),
+        })
+    }
+
+    #[test]
+    fn create() {
+        let client = test_client();
+        let pool = client.rocket().state::<DbPool>().expect("DbPool is not managed");
+        let conn = DbConn(pool.get().expect("Test DB pool error"));
+        setup_db(&*conn);
```
fdb-hiroshima: anything after `let conn = ...` should be surrounded by a big
```rust
conn.test_transaction::<_, (), _>(|| { 
    [...]
    Ok(())
})
```
to prevent any side effect

fdb-hiroshima:
Is this still ready for review? If so you should fix the conflict with master

PR #309, opened 2018-11-06
Test with PostgreSQL too

BaptisteGelez:
The coverage is still only reported with SQLite, but it should be the same anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ cache:
   - kcov-master
 sudo: true
 dist: trusty
+services: postgres
 
 addons:
   apt:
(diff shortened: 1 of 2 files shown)


fdb-hiroshima:
You can probably add 'test' to the stages, so they are run in a logical order (test before code cov), or maybe we should calculate code coverage with both postgres and sqlite, and change the name of the `code coverage` stage, so it's more symmetrical

BaptisteGelez:
I don't know how Codecov will react if we send two different coverage reports for the same commit…

fdb-hiroshima:
They "add" the two report, a line caught by at least one test is considered covered, lines caught by no tests are considered missed

BaptisteGelez:
OK, great! I will try to get coverage for Postgres too then.

BaptisteGelez:
Oh thanks for your commit!

PR #310, opened 2018-11-07
Add unit tests for main model parts

fdb-hiroshima:
Add tests for following model parts:
- [x] Blog
- [ ] Comment
- [x] Instance
- [x] Media
- [ ] Post
- [x] User


fdb-hiroshima:
I think some part of the code should be refactored before going further. Database initialization is getting messier and messier because some things currently done in routes should be done in the model. I'll start again latter when those things will have moved

fdb-hiroshima:
I've rebased to master to allow for merge

Issue #273, opened 2018-10-09
Change the logo from configuration


BaptisteGelez:
We should have all the submitted logos in the repository, and the possibility to change the one that will be displayed with a `PLUME_LOGO` (or something similar) environment variable.

Plume version (and/or commit): 0.2.0

BaptisteGelez:
Marking this as blocked until #161 is closed.

trwnh:
I was thinking more along the line of inlining the logo directly into the HTML, instead of simply linking to it? In other words, allow admins to dynamically select a PNG or an SVG, and if it is a PNG, then wrap it in an `<img>` tag, and if it is an SVG, then include it directly (after sanitizing it and adding a class/id).

An example I've made: https://github.com/trwnh/plumeLogos/blob/master/sampleSVG.html uses plumeFeather.min.svg and plumeParagraphs.min.svg directly within the HTML, which allows for styling the logo with CSS :)

Specifically, an admin panel would allow admins to set the size, fill, background, and even the stroke width as desired!

BaptisteGelez:
Yeah, that would be great, and it could be possible to either include SVG or an <img> tag with askama.

fdb-hiroshima:
We use more than one logo (filled or not), and the manifest.json advertise many different picture sizes. How should we handle that with custom logos?
My best idea at the moment would be to ask for a `PLUME_LOGO`, which disable all builtin, and an optional `PLUME_LOGO_FAVICON` for the second logo, which default to the first if none provided, as well as reading any `PLUME_LOGO_{int}` and use them for the manifest.json

BaptisteGelez:
@fdb-hiroshima Sounds good to me. :+1:

Issue #110, opened 2018-07-06
Implement url_for and use rendered links


bitkeks:
Currently Plume templates use hardcoded endpoint URLs. This can break, like mentioned in #65 and might be a problem later if routes change. We should therefore use a function like `url_for` in Flask to render endpoint URLs from a map of routes.

PR #327, opened 2018-11-29
Use Ructe

### LINKED ISSUE #161: Consider using askama for templates
[Askama](https://github.com/djc/askama) is a templating library that compiles templates into native Rust code. It means that templates are type-checked, and that you can use macros like `uri!` inside of them, which could be really useful.

BaptisteGelez:
All the template are now compiled at compile-time with the `ructe` crate.

I preferred to use it instead of askama because it allows more complex Rust expressions, where askama only supports a small subset of expressions and doesn't allow them everywhere (for instance, `{{ macro!() | filter }}` would result in a parsing error).

The diff is quite huge, but there is normally no changes in functionality.

Fixes #161 and unblocks #110 and #273


BaptisteGelez:
It also updates Rocket to 0.4 RC 1, and Rocket i18n to the latest commits (that I should push btw, I completely forgot :sweat_smile:).

Cargo.toml
```diff
@@ -49,26 +50,18 @@ path = "plume-common"
...
 git = "https://github.com/fdb-hiroshima/rocket_csrf"
-rev = "2805ce5dbae4a6441208484426440885a5640a6a"
+rev = "e6ead8051cab12a95ea9dc88db3a107f0b5f7533"
 
 [dependencies.rocket_i18n]
-git = "https://github.com/BaptisteGelez/rocket_i18n"
-rev = "75a3bfd7b847324c078a355a7f101f8241a9f59b"
+# git = "https://github.com/BaptisteGelez/rocket_i18n"
+# rev = "f8f9913f6f7008f65ae155dcd0e730377db6edae"
+path = "../rocket_i18n"
```
fdb-hiroshima: I knew it would land to the pr x)

BaptisteGelez: Yes, I just remembered I forgot it. :grin:

src/api/mod.rs
```diff
@@ -10,16 +11,16 @@ use plume_models::{
...
     client_id: String,
     client_secret: String,
     password: String,
     username: String,
     scopes: String,
 }
 
-#[get("/oauth2?<query>")]
-fn oauth(query: OAuthRequest, conn: DbConn) -> Json<serde_json::Value> {
+#[get("/oauth2?<query..>")]
```
fdb-hiroshima: what is the `..` in a query part for? I only know its use in the path part

src/routes/mod.rs
```diff
@@ -1,67 +1,30 @@
...
-        write!(f, "page={}", &self.page as &UriDisplay)
-    }
-}
+pub struct Page(i32);
 
-impl FromUriParam<i32> for Page {
-    type Target = Page;
-    fn from_uri_param(num: i32) -> Page {
-        Page { page: num }
+impl<'v> FromFormValue<'v> for Page {
```
fdb-hiroshima: ugh I did the same change on `search`, we will have some merge issues for literally nothing :cry: (and also in a lot of other places, but for good reasons)

src/routes/mod.rs
```diff
@@ -97,6 +60,26 @@ pub fn post_to_atom(post: Post, conn: &Connection) -> Entry {
         .build().expect("Atom feed: entry error")
 }
 
+type Ructe = Content<Vec<u8>>;
+
+macro_rules! render {
```
fdb-hiroshima: It would feel more at it's place in `src/template_utils.rs` I think

src/template_utils.rs
```diff
@@ -0,0 +1,126 @@
...
+    fn as_str(&self) -> &'static str {
+        match self {
+            Size::Small => "small",
+            Size::Medium => "medium",
+        }
+    }
+}
+
+pub fn avatar(conn: &Connection, user: &User, size: Size, pad: bool, catalog: &Catalog) -> Html<String> {
+    Html(format!(
```
fdb-hiroshima: I haven't tested yet, but big warning on fields like `title`, one could maybe use a name containing html (or js) and get it unescaped in his page

po/fr.po
```diff
@@ -259,21 +258,21 @@ msgstr "Vous devez vous connecter pour suivre quelqu’un"
 msgid "You need to be logged in order to edit your profile"
 msgstr "Vous devez vous connecter pour modifier votre profil"
 
-msgid "By {{ link_1 }}{{ link_2 }}{{ link_3 }}{{ name | escape }}{{ link_4 }}"
-msgstr ""
-"Par {{ link_1 }}{{ link_2 }}{{ link_3 }}{{ name | escape }}{{ link_4 }}"
+#, fuzzy
+msgid "By {0}"
```
fdb-hiroshima: is it normal for `{0}` to be translated by `{0}{1}{2}`?

plume-common/src/utils.rs
```diff
@@ -23,8 +22,13 @@ pub fn make_actor_id(name: &str) -> String {
...
 }
 
+/**
+* Redirects to the login page with a given message.
+*
+* Note that the message should be translated before passed to this function.
+*/
 pub fn requires_login<T: Into<Uri<'static>>>(message: &str, url: T) -> Flash<Redirect> {
-    Flash::new(Redirect::to(format!("/login?m={}", gettext(message.to_string()))), "callback", url.into().to_string())
+    Flash::new(Redirect::to(format!("/login?m={}", message)), "callback", url.into().to_string())
```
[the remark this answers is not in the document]
BaptisteGelez: You are right, I forgot to call it from the routes (I preferred to change this behavior to avoid having to depend on rocket_i18n in plume-common, and to have an extra `Catalog` parameter).

templates/partials/post_card.rs.html
```diff
@@ -0,0 +1,29 @@
...
+<div class="card">
+    @if article.cover_id.is_some() {
+        <div class="cover" style="background-image: url('@Html(article.cover_url(ctx.0).unwrap_or_default())')"></div>
+    }
+    <h3><a href="@article.url(ctx.0)">@article.title</a></h3>
+    <main>
+        <p>@article.subtitle</p>
+    </main>
+    <p class="author">
+        @Html(i18n!(ctx.1, "By {0}"; format!(
```
fdb-hiroshima: there is an xss injection there, because {0} is not properly escaped. Example paylod: display name = `"><script>alert("xss");</script> <div title="`

src/template_utils.rs
fdb-hiroshima: I was right on this one `"><script>alert("xss");</script> <div title="` as a display name trigger it very

src/api/mod.rs
BaptisteGelez: It is the new way to handle queries in Rocket 0.4. The new syntax is:

- `/?<x>` is a single query argument (`?x=foo`)
- You can separate query arguments with `&` (`/?<x>&<y>` will match `?x=foo&y=bar`)
- If you want to deserialize a whole struct, you have to use `..` and each of its field will be a query argument (I think it was the default behavior in 0.3)

po/fr.po
BaptisteGelez: No, it is a translation I forgot to update I think.

templates/partials/post_card.rs.html
BaptisteGelez: Yes, I guess we could use `.to_html` from Ructe to escape it https://github.com/kaj/ructe/blob/master/src/template_utils.rs#L24

plume-models/src/medias.rs
```diff
@@ -47,65 +54,65 @@ impl Media {
...
-        json["html"] = json!(html);
-        json["url"] = json!(url);
-        json["md"] = json!(md);
-        json["category"] = json!(cat);
-        json
+            MediaCategory::Unknown => String::new(),
+        }
+    }
+
+    pub fn html(&self, conn: &Connection) -> String {
```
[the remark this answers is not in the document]
fdb-hiroshima: The same payload is also working, however as this is only used on private pages, this has less impact (one can only harm themselves)

templates/partials/post_card.rs.html
fdb-hiroshima: then it should be called only on L18, as to not break the link we build

BaptisteGelez: Would doing that break anything?

```rust
i18n!(ctx.1, "By {0}"; format!(
    "<a href=\"/@/{}/\">{}</a>",
    article.get_authors(ctx.0)[0].get_fqn(ctx.0).to_html(),
    article.get_authors(ctx.0)[0].name(ctx.0).to_html()
))
```

plume-common/src/utils.rs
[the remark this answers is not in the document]
fdb-hiroshima: I totally understand your thought, it will be more logical to do so

templates/instance/users.rs.html
```diff
@@ -0,0 +1,33 @@
...
+    <div class="list">
+        @for user in users {
+            <div class="flex">
+                @avatar(ctx.0, &user, Size::Small, false, ctx.1)
+                <p class="grow">
+                    <a href="/@@/@user.get_fqn(ctx.0)">@user.name(ctx.0)</a>
+                    <small>@format!("@{}", user.username)</small>
+                </p>
+                @if !user.is_admin {
+                    <form class="inline" method="post" href="/admin/users/@user.id/ban">
```
fdb-hiroshima: The attribute to use is target, not href (this issue was introduced on Tera)
```suggestion
                    <form class="inline" method="post" target="/admin/users/@user.id/ban">
```

templates/blogs/details.rs.html
```diff
@@ -0,0 +1,49 @@
...
+    <section>
+        <h2>
+            @i18n!(ctx.1, "Latest articles")
+            <small><a href="/~/@fqn/atom.xml" title="Atom feed">@icon!("rss")</a></small>
+        </h2>
+        @if posts.len() < 1 {
+            <p>@i18n!(ctx.1, "No posts to see here yet.")</p>
+        }
+        @if is_author {
+            <a href="/~/@fqn/new/" class="button inline-block">@i18n!(ctx.1, "New article")</a>
```
fdb-hiroshima: I'm not sure why yet, but having a user&display name of `" content=` shift the button to the left

templates/instance/about.rs.html
```diff
@@ -0,0 +1,40 @@
...
+          </div>
+          <div>
+              <p>@i18n!(ctx.1, "And connected to")</p>
+              <em>@n_instances</em>
+              <p>@i18n!(ctx.1, "other instances")</p>
+          </div>
+          <div>
+              <p>@i18n!(ctx.1, "Administred by")</p>
+              @avatar(ctx.0, &admin, Size::Small, false, ctx.1)
+              <p><a href="/@@/@admin.get_fqn(ctx.0)">@admin.name(ctx.0)</a><small>@format!("@{}", admin.get_fqn(ctx.0))</small></p>
```
fdb-hiroshima: You can do without `format` (not sure if it's better though)
```suggestion
              <p><a href="/@@/@admin.get_fqn(ctx.0)">@admin.name(ctx.0)</a><small>@@@admin.get_fqn(ctx.0)</small></p>
```

templates/medias/details.rs.html
```diff
@@ -0,0 +1,34 @@
...
+        </figure>
+        <div>
+            <p>
+                @i18n!(ctx.1, "Markdown code")
+                <small>@i18n!(ctx.1, "Copy it in your articles to insert this media.")</small>
+            </p>
+            <code>@media.markdown(ctx.0)</code>
+        </div>
+        <div>
+            <form class="inline" method="post" action="/medias/@media.id/avatar">
```
fdb-hiroshima: this button should not be here if MediaCategory != Image
(was true for Tera too though)

templates/partials/instance_description.rs.html
```diff
@@ -0,0 +1,37 @@
...
+        </div>
+        <div class="presentation card">
+            <h2>@i18n!(ctx.1, "About {0}"; instance.name)</h2>
+            <main>
+                @Html(instance.short_description_html)
+                <section class="stats">
+                    <div>
+                        <p>@i18n!(ctx.1, "Home to")</p>
+                        <em>@n_users</em>
+                        <p>@i18n!(ctx.1, "people")</p>
```
fdb-hiroshima: same about translating with a single string

templates/posts/details.rs.html
```diff
@@ -0,0 +1,146 @@
...
+                </form>
+            </div>
+        } else {
+            <p class="center">@i18n!(ctx.1, "Login or use your Fediverse account to interact with this article")</p>
+            <div class="actions">
+                <div class="likes">
+                    <p aria-label="@i18n!(ctx.1, "One like", "{0} likes", &n_likes)" title="@i18n!(ctx.1, "One like", "{0} likes", n_likes)">
+                        @n_likes
+                    </p>
+                    <a href="/login?m=Login%20to%20like" class="action">@icon!("heart") @i18n!(ctx.1, "Add yours")</a>
```
fdb-hiroshima: Message won't get translated

templates/posts/details.rs.html
```diff
@@ -0,0 +1,146 @@
...
+                </div>
+
+                <div class="reshares">
+                    <p aria-label="@i18n!(ctx.1, "One boost", "{0} boost", &n_reshares)" title="@i18n!(ctx.1, "One boost", "{0} boosts", n_reshares)">
+                        @n_reshares
+                    </p>
+                    <a href="/login?m=Login%20to%20boost" class="action">@icon!("repeat") @i18n!(ctx.1, "Boost")</a>
```
fdb-hiroshima: Message won't get translated

templates/users/details.rs.html
```diff
@@ -0,0 +1,36 @@
...
+@use plume_models::posts::Post;
+@use plume_models::users::User;
+
+@(ctx: BaseContext, user: User, follows: bool, is_remote: bool, remote_url: String, recents: Vec<Post>, reshares: Vec<Post>)
+
+@:base(ctx, &user.name(ctx.0), {}, {}, {
+    @:header(ctx, &user, follows, is_remote, remote_url)
+
+    @tabs(&[
+        (&format!("/@/{}", user.get_fqn(ctx.0)), i18n!(ctx.1, "Articles"), true),
```
fdb-hiroshima: user fqn can't be assumed to be trusted yet, one can inject content by this way

templates/users/followers.rs.html
```diff
@@ -0,0 +1,24 @@
...
+@use template_utils::*;
+@use plume_models::users::User;
+
+@(ctx: BaseContext, user: User, follows: bool, is_remote: bool, remote_url: String, followers: Vec<User>, page: i32, n_pages: i32)
+
+@:base(ctx, &i18n!(ctx.1, "{0}'s followers"; user.name(ctx.0)), {}, {}, {
+    @:header(ctx, &user, follows, is_remote, remote_url)
+
+    @tabs(&[
+        (&format!("/@/{}", user.get_fqn(ctx.0)), i18n!(ctx.1, "Articles"), false),
```
fdb-hiroshima: same here, user fqn can't be trusted yet

src/template_utils.rs
```diff
@@ -0,0 +1,151 @@
...
+        url = user.avatar_url(conn),
+        title = i18n!(catalog, "{0}'s avatar"; name),
+    ))
+}
+
+pub fn tabs(links: &[(&str, &str, bool)]) -> Html<String> {
+    let mut res = String::from(r#"<div class="tabs">"#);
+    for (url, title, selected) in links {
+        res.push_str(r#"<a href=""#);
+        res.push_str(url);
```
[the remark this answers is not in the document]
fdb-hiroshima: fixed by 9714bafded2aefeaf2eb4123a338b93c4cd019d8

templates/blogs/details.rs.html
fdb-hiroshima: this was only on user name (fqn), fixed by 9714bafded2aefeaf2eb4123a338b93c4cd019d8 then

templates/users/details.rs.html
fdb-hiroshima: fixed by 9714bafded2aefeaf2eb4123a338b93c4cd019d8

templates/users/followers.rs.html
fdb-hiroshima: fixed by 9714bafded2aefeaf2eb4123a338b93c4cd019d8

templates/posts/details.rs.html
fdb-hiroshima: actually as there are route for it, and has_liked and has_reshared are both set to false by default, the whole `@if ctx.2.is_some() {} else {}` could be turned into a single branch. That way there is less code and it work better. The only issue is there would be no csrf token (no session cookie so rocket_csrf don't issue a token), so the user would be faced with an invalid csrf page. I should probably make rocket_csrf accept request with no cookie, not only response. Anyway it make no sens to do a csrf attack on a client having no session. And this would also remove most exceptions we set in main

templates/partials/post_card.rs.html
```diff
@@ -0,0 +1,29 @@
+@use template_utils::*;
+@use plume_models::posts::Post;
+
+@(ctx: BaseContext, article: Post)
+
+<div class="card">
+    @if article.cover_id.is_some() {
+        <div class="cover" style="background-image: url('@Html(article.cover_url(ctx.0).unwrap_or_default())')"></div>
```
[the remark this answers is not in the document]
fdb-hiroshima: Fixed by ed71d24fe9bb66244474c87959c07dd422baeb14

src/template_utils.rs
```diff
@@ -0,0 +1,151 @@
...
+    Html(format!(
+        r#"<div
+        class="avatar {size} {padded}"
+        style="background-image: url('{url}');"
+        title="{title}"
+        aria-label="{title}"
+        ></div>"#,
+        size = size.as_str(),
+        padded = if pad { "padded" } else { "" },
+        url = user.avatar_url(conn),
```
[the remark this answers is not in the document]
fdb-hiroshima: fixed by ed71d24fe9bb66244474c87959c07dd422baeb14

plume-models/src/medias.rs
```diff
@@ -47,65 +56,65 @@ impl Media {
...
+            "mp4" | "avi" | "webm" | "mov" => MediaCategory::Video,
+            _ => MediaCategory::Unknown,
+        }
+    }
+
+    pub fn preview_html(&self, conn: &Connection) -> SafeString {
+        let url = self.url(conn);
+        match self.category() {
+            MediaCategory::Image => SafeString::new(&format!(
+                "<img src=\"{}\" alt=\"{}\" title=\"{}\" class=\"preview\">",
```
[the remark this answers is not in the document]
fdb-hiroshima: @igalic  because I always do this mistake, warning on the end of your string, the correct syntax is `r#"some content"#`

templates/errors/base.rs.html
```diff
@@ -0,0 +1,9 @@
+@use templates::base as base_template;
+@use template_utils::*;
+
+@(ctx: BaseContext, error_message: &str, error: Content)
+
+@:base_template(ctx, error_message, {}, {}, {
+  @:error()
```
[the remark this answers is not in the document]
BaptisteGelez: It comes from the `error` parameter of this template. The `Content` type is a special type used by Ructe to pass template fragments to functions. Then, to inherit a template, you can call it (as we are doing with `@:base_template` here) and give it fragments of templates to be included as arguments between `{}`.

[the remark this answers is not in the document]
fdb-hiroshima: How ok I through it was a function call when it's more of a "insert the content of this var here"

po/LINGUAS
```diff
@@ -1,8 +0,0 @@
-en
-fr
-pl
```
fdb-hiroshima: I think this is easier for translators to add new languages in here than it is to edit main.rs. Maybe include_i18! should read this file instead of getting languages from a static parameter (actually it could accept both), what do you think @BaptisteGelez ?

BaptisteGelez: Ideally, I think include_i18n! should read all the .mo files in `translations` if no language list is provided. This way new languages would be loaded automatically, without needing to add them anywhere.

fdb-hiroshima: It seems like it's not possible to `ls` with normal macros, however it exist a crate doing more or less what we want [here](https://github.com/pyros2097/rust-embed) (we could also use it to embed wasm files later on)

fdb-hiroshima: I think using any of those isn't in the range of this pr anyway, a macro is a thing, a dependency is another

po/ja.po
```diff
@@ -0,0 +1,601 @@
+msgid ""
```
fdb-hiroshima: you forgot to add it to include_i18n

PR #332, opened 2018-12-02
Upgrade plume dependencies

fdb-hiroshima:
Not much changed, I just changed some code because of validator 0.8.0 breaking changes, and removed hyper from plume-common dependencies as it was somehow unnecessary


Cargo.toml
```diff
@@ -5,20 +5,20 @@ version = "0.2.0"
...
-canapi = "0.1"
+canapi = "0.2"
 colored = "1.6"
 dotenv = "0.13"
 failure = "0.1"
 gettext-rs = "0.4"
 guid-create = "0.1"
 heck = "0.3.0"
-rpassword = "2.0"
+rpassword = "2.1"
```
BaptisteGelez: I think we don't need rpassword anymore in Plume itself (only in the CLI), maybe we should remove it with this PR too?

fdb-hiroshima: I did it for hyper, there is no reason to not do it for rpassword too

fdb-hiroshima:
I'm just checking everything is fine, but compiling with a new rustc require to rebuild all dependencies

Issue #241, opened 2018-09-20
Feature request: single user mode


jeroenpraat:
Right now Plume is aimed towards multiple users on one instance. It would be great if there's a single user mode.

This would actually mean that the root of the (sub)domain (https://blog.example.com) would show a specific blog instead of Plume's landing page.

Plume version (and/or commit): 0.2.0

BaptisteGelez:
Hello!

I consider this issue a duplicate of #94, please re-open if what the implementation I proposed there is not correctly covering your use case.

Issue #94, opened 2018-06-28
Serve multiple domains on a single instance


DanielHeath:
I'd love to be able to run this for a few friends - they just have to point their DNS at my VPS.

That way, users can have a hosted option, but also keep the ability to move to another host easily / host their own in future.

BaptisteGelez:
We are currently discussing about this feature here: https://framavox.org/d/qF3vNcfw/custom-blog-domains-and-where-to-attach-them (do we really need it, and if yes, how should it be implemented?)

igalic:
i'll try to summarize the outcome of the discussion 

× | instance | blog | author
-- | -- | -- | --
| *domain* | always | same | same
| *custom_domain* | No | Yes | No
| *URLs* | https://blog.eena.me/ | https://blog.eena.me/~/words/ | https://blog.eena.me/@/meena
| | | https://blog.eena.me/~/thoughts/ | |
| *Custom URL* | × | https://words.eena.me/ | ×
| | × | https://thoughts.eena.me/ | ×

to piece together a better idea on how to implement it.

BaptisteGelez:
So, to summarize what we have here. We want to add an option to have custom domains for **blogs** (not users or anything else). In the database this will be materialized by a new nullable/optional `custom_domain` field on `blogs` or something like that.

A question that we still need to discuss, and that will have a quite big impact on how this feature is implemented: do we allow only sub-domains of the main domain, only totally custom domains (owned by the same person as the blog), or both? In the first case, most of the configuration will be done by the instance admin: they will have to redirect all subdomains (is it possible to use wildcards in DNS records, or will them have to add each domain manually (which would make this feature far less interesting in my opinion)?) to the correct IP, and configure their reverse-proxy to make the communication between these domains and  Plume. In the second case, it will be bloggers' job to configure the DNS, but what about the reverse-proxy? What we allow will also have an impact on the UI to configure custom domains (in one case, you just check something and you are assigned a custom domain, in the other you have to tell Plume what your domain is).

Independently of what of domains we allow, it means that Plume will have to act a little bit like a reverse proxy itself, displaying different pages depending on the domain. To make it easier to implement this feature, I think we should only display a few pages (the homepage with the blog details and the page to read an article basically) under the custom domain. To know if the current route should be interpreted in a general context, or as a custom domain route, I think implementing a request guard that returns the `Host` header if it is different from `BASE_URL` could help. This way you could do:

```rust
#[get("/")]
[...]

PR #340, opened 2018-12-08
add a custom_domain attribute for single-user blogs

igalic:
We want to be able to serve a blog from a custom domain

this PR tries to follow what @BaptisteGelez's describes in #241 #94


plume-models/src/schema.rs
```diff
@@ -183,6 +183,7 @@ table! {
         is_admin -> Bool,
         summary -> Text,
         email -> Nullable<Text>,
+        custom_domain -> Nullable<Text>,
```
fdb-hiroshima: Why is custom_domain `NOT NULL` in migrations but `Nullable` here?

templates/users/new.rs.html
```diff
@@ -11,6 +11,7 @@ <h1>@i18n!(ctx.1, "Create an account")</h1>
         <form method="post">
             @input!(ctx.1, username (text), "Username", form, errors.clone(), "minlenght=\"1\"")
             @input!(ctx.1, email (text), "Email", form, errors.clone())
+            @input!(ctx.1, email (text), "Email", form, errors.clone(), "")
```
BaptisteGelez: I think we shouldn't ask for a custom domain when registering (I guess that's what you wanted to do here, even if it looks like you forgot to update what you copied)

igalic: thanks for the catch, i changed this to:

```rust
            @input!(ctx.1, custom_domain (optional text), "custom_domain", form, errors.clone(), "")
```

and now i get this joyful error:

```
   --> src/template_utils.rs:127:30                                                                                                                                           
    |                                                                                                                                                                         
127 |                 val = escape(&$form.$name),                                                                                                                             
    |                              ^^^^^^ expected str, found enum `std::option::Option`                                                                                      
    |                                                                                                                                                                         
   ::: /home/meena/src/ap/Plume/target/debug/build/plume-7e8607967e0d30ae/out/templates/users/template_new.rs:21:1                                                            
    |                                                                                                                                                                         
21  | input!(ctx.1, custom_domain (optional text), "custom_domain", form, errors.clone(), "").to_html(out)?;                                                                  
    | --------------------------------------------------------------------------------------- in this macro invocation                                                        
    |                                                                                                                                                                         
    = note: expected type `&str`                                                                                                                                              
               found type `&std::option::Option<std::string::String>`                                                                                                         
                                                                                                                                                                              
error: aborting due to previous error                                                                                                                                         
                                                                                                                                                                              
For more information about this error, try `rustc --explain E0308`.                                                                                                           
error: Could not compile `plume`.                                                                                                                                             
```

fdb-hiroshima: input! don't support `Option`, make `custom_domain` (in `NewUserForm`) not optional, and map it to None whenever it's empty in `create(...)`

plume-cli/src/users.rs
```diff
@@ -59,6 +64,7 @@ fn new<'a>(args: &ArgMatches<'a>, conn: &Connection) {
     let admin = args.is_present("admin");
     let bio = args.value_of("biography").unwrap_or("").to_string();
     let email = args.value_of("email").map(String::from).unwrap_or_else(|| super::ask_for("Email address"));
+    let custom_domain = args.value_of("custom-domain").unwrap_or("").to_string();
```
fdb-hiroshima: I think you shouldn't unwrap. If it's None, then create the new user with custom_domain = None, otherwise a user with no custom domain would be stored as having the custom domain "" (empty string)

igalic: @fdb-hiroshima do we just `.clone()` here?

fdb-hiroshima: I think I would do
```rust
    let custom_domain = args.value_of("custom-domain").map(str::to_owned);
```

igalic: thanks!

migrations/postgres/2018-12-08-191238_users_add_custom_domain/up.sql
```diff
@@ -0,0 +1,2 @@
+-- Your SQL goes here
+ALTER TABLE users ADD COLUMN custom_domain VARCHAR NOT NULL DEFAULT '';
```
BaptisteGelez: Maybe it should be unique?

igalic: that would heavily imply, in code, that we're taking #94 seriously, in that the instance then is single-user, eh?

migrations/sqlite/2018-12-08-191758_users_add_custom_domain/down.sql
```diff
@@ -0,0 +1,31 @@
+-- This file should undo anything in `up.sql`
+PRAGMA foreign_keys = ON;
+CREATE TABLE migrate_users_custom_domain (
```
igalic: i realized we can give our temporary tables sensible names!

igalic: i realized we can give our temporary tables sensible names!

migrations/postgres/2018-12-08-191238_users_add_custom_domain/up.sql
BaptisteGelez: Hum… yes, maybe it makes sense to have the same domain for many users.

plume-models/src/schema.rs
igalic: does this mean we can resolve this?

fdb-hiroshima: You can either leave this open or resolve and mark [this](https://github.com/Plume-org/Plume/pull/340/files#diff-5b7775a42e5a98eee192a505e07337baR2) and [that](https://github.com/Plume-org/Plume/pull/340/files#diff-03a9c6038bd075d7cb69055575b8f298R2) instead

igalic:
With the conclusions that emerged from [the discussion (Custom Blog domains and where to attach them)](https://framavox.org/d/qF3vNcfw/custom-blog-domains-and-where-to-attach-them) i've decided to close this, and with a better understanding, open a new pull request.

Thank you all for helping me with this!

Issue #113, opened 2018-07-10
Some comments are duplicated


BaptisteGelez:
If you look at https://baptiste.gelez.xyz/~/PlumeDevelopment/this-month-in-plume-june-2018/ it seems to happen especially on comments with multiple mentions.

BaptisteGelez:
May be related to #80 too

ffffluffy:
It seems I'm one of the users affected by this duplication bug.

Comment without mention :
  * https://baptiste.gelez.xyz/~/Announcements@blog.funkwhale.audio/funkwhale-past-present-and-future/
  * https://baptiste.gelez.xyz/~/Testmiraty/ben-voila/
  * https://baptiste.gelez.xyz/~/FluffysTests/this-is-a-media-test/
  * https://baptiste.gelez.xyz/~/FluffysAdventures/writing-prompts-the-desert's-shadow/

Comment with one mention :
  * https://baptiste.gelez.xyz/~/Plume/plume/
  * Here, it affects also @ 0x1C3B00DA https://baptiste.gelez.xyz/~/MakeBlogPhoe/how-does-one-use-activity-pub-for-games/

However, here for example, I don't have any duplication : https://baptiste.gelez.xyz/~/Nguy%C3%AAn/test-blog-post-1/

BaptisteGelez:
Thanks for investigating this!

BaptisteGelez:
I think it happens when someone from the same Plume instance follows you.

fdb-hiroshima:
There are still issues with duplicated comments [here](https://baptiste.gelez.xyz/~/PlumeDevelopment/plume's-first-release/#comment-502). The comment appear 3 times, and conveniently that's also the number of peoples from the Plume instance who got mentioned. However this comment was originated from Mastodon

BaptisteGelez:
@fdb-hiroshima I guess Mastodon sent it to each users' inbox, running thrice the creation code in Plume.

fdb-hiroshima:
Closed by #342

PR #342, opened 2018-12-08
Add some constraint at database level

### LINKED ISSUE #79: Add a uniqueness constraint on the ap_url field

### LINKED ISSUE #201: Duplicated instances
I was surprised at the number of federated instances : 
screenshot-2018-9-6 about le blog d un e enby anar geek plume

In the database, some instances are duplicated:

 `SELECT * FROM "instances"`
screenshot-2018-9-6 selectionner instances - 127 0 0 1 - adminer

BaptisteGelez:
Fixes #79 and should fix #201 and #113 as well

- The `value` of `ApiToken` is unique
- A blog/user with a given ID is unique on its instance
- An author can't be twice in a blog or post
- AP-related URLs/IDs are unique
- Instance domains are unique
- Like/reshare by an user on a specific post are unique
- Media have an unique path
- A tag can't be inserted twice in the same post (actually it can, but once as a hashtag, once as a regular tag)


fdb-hiroshima:
There is some issue with migrating on an instance that don't currently met the constrain, I'm getting a `could not create unique index "blog_unique"` because I've some duplicated blogs (I think those are old tests leaks in my case, but as some will have some constraint violated, we should make sure the migration goes smoothly)

migrations/postgres/2018-12-08-175515_constraints/up.sql
```diff
@@ -0,0 +1,25 @@
...
+ALTER TABLE blogs ADD CONSTRAINT blog_unique UNIQUE (actor_id, instance_id);
+ALTER TABLE blogs ADD CONSTRAINT blog_unique_ap_url UNIQUE (ap_url);
+ALTER TABLE blogs ADD CONSTRAINT blog_unique_outbox UNIQUE (outbox_url);
+ALTER TABLE blogs ADD CONSTRAINT blog_unique_inbox UNIQUE (inbox_url);
+ALTER TABLE comments ADD CONSTRAINT comments_unique_ap_url UNIQUE (ap_url);
+ALTER TABLE follows ADD CONSTRAINT follows_unique_ap_url UNIQUE (ap_url);
+ALTER TABLE instances ADD CONSTRAINT instance_unique_domain UNIQUE (public_domain);
+ALTER TABLE likes ADD CONSTRAINT likes_unique UNIQUE (user_id, post_id);
+ALTER TABLE likes ADD CONSTRAINT likes_unique_ap_url UNIQUE (ap_url);
+ALTER TABLE medias ADD CONSTRAINT medias_unique_path UNIQUE (file_path);
```
fdb-hiroshima: This constrain seems to forbid from having more than one remote medias (it allows for only one null)

migrations/postgres/2018-12-08-175515_constraints/down.sql
```diff
@@ -0,0 +1,25 @@
...
+ALTER TABLE blogs DROP CONSTRAINT IF EXISTS blog_unique;
+ALTER TABLE blogs DROP CONSTRAINT IF EXISTS blog_unique_ap_url;
+ALTER TABLE blogs DROP CONSTRAINT IF EXISTS blog_unique_outbox;
+ALTER TABLE blogs DROP CONSTRAINT IF EXISTS blog_unique_inbox;
+ALTER TABLE comments DROP CONSTRAINT IF EXISTS comments_unique_ap_url;
+ALTER TABLE follows DROP CONSTRAINT IF EXISTS follows_unique_ap_url;
+ALTER TABLE instances DROP CONSTRAINT IF EXISTS instance_unique_domain;
+ALTER TABLE likes DROP CONSTRAINT IF EXISTS likes_unique;
+ALTER TABLE likes DROP CONSTRAINT IF EXISTS likes_unique_ap_url;
+ALTER TABLE medias DROP CONSTRAINT IF EXISTS medias_unique_path;
```
fdb-hiroshima: this constraint won't exist (won't fail because drop if exists, but also useless)

PR #343, opened 2018-12-09
License federation

### LINKED ISSUE #259: Actually federate license
We currently fallback to the default license for remote articles, but it should be federated. I think Peertube federates it, we should see how it is done there.

BaptisteGelez:
Fixes #259 

I decided not to follow what Peertube was doing, and to use a simple `license` field with the name of the license, because their system is not flexible enough IMO, each AP implementation would have to keep the same list of licenses, and it wouldn't allow any other license outside of that.

A bit unrelated, but I also made it possible to have no specific license to keep all your rights on what you wrote. The license field will now be pre-filled with the default license when creating a post, and emptying it will display "All rights reserved" instead of the name of the license.

### DIFF
--- a/plume-common/src/activity_pub/mod.rs
+++ b/plume-common/src/activity_pub/mod.rs
@@ -220,3 +220,12 @@ pub struct Source {
 }
 
 impl Object for Source {}
+
+#[derive(Clone, Debug, Default, Deserialize, Serialize, Properties)]
+#[serde(rename_all = "camelCase")]
+pub struct Licensed {
+    #[activitystreams(concrete(String), functional)]
+    pub license: Option<serde_json::Value>,
+}
+
+impl Object for Licensed {}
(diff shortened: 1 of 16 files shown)


BaptisteGelez:
What if we `.unwrap_or(instance_default_license)` here: https://github.com/Plume-org/Plume/pull/343/files#diff-1b633efbc0716f65e55e3170eb127764R796 ? Maybe this behavior would be better? Or is it worse?

fdb-hiroshima:
I don't know, saying all right reserved is maybe more conservative. Telling peoples "this is CC-BY-SA" when it should be all right reserved seems to me like a bigger problem than the opposite, "all right reserved" might be false but at least it won't induce people into violating licenses involuntarily. Like most of the time when you simply have not enough information, no solution is perfect :neutral_face:

PR #344, opened 2018-12-09
Count items in database as much as possible

BaptisteGelez:


plume-models/src/notifications.rs
```diff
@@ -48,6 +48,15 @@ impl Notification {
             .expect("Notification::find_for_user: notification loading error")
     }
 
+    pub fn count_for_user(conn: &Connection, user: &User) -> i64 {
+        notifications::table
+            .filter(notifications::user_id.eq(user.id))
+            .order_by(notifications::creation_date.desc())
```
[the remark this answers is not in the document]
BaptisteGelez: Yes, I just copy/pasted another function. I'll change that.

plume-models/src/blogs.rs
```diff
@@ -83,6 +83,19 @@ impl Blog {
...
     }
 
+    pub fn count_authors(&self, conn: &Connection) -> i64 {
+        use schema::blog_authors;
+        use schema::users;
+        let authors_ids = blog_authors::table
+            .filter(blog_authors::blog_id.eq(self.id))
+            .select(blog_authors::author_id);
+        users::table
+            .filter(users::id.eq_any(authors_ids))
```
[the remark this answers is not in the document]
BaptisteGelez: OK, I'll change it (and yes you were clear enough :slightly_smiling_face:)

igalic:
> Love it when an issue is only reproducible on one db backend. It might be because sqlite don't enforce some constraints by default, we could enabled `PRAGMA foreign_keys = ON;`, but I wonder what changes to make this test fail

we have that enabled — on the `users` table — not sure if/how we enable it by default.

fdb-hiroshima:
@igalic I think we only set it in a migration, this property is not stored so we need to set it somehow at each Plume launch, and this is a global var, so it was enabled for all tables (but only until the migration end)

igalic:
> this property is not stored

unlike (some) other `pragma` instructions, it's not storable… it's, as you say, per connection or per transaction.

So perhaps the diesel sqlite driver has something we can turn on by default
------
i checked, it can't
but maybe ***we*** can wrap the query functions with a `begin transaction; pragma foreign_keys = on;`

plume-models/src/db_conn.rs
```diff
@@ -38,3 +38,15 @@ impl Deref for DbConn {
         &self.0
     }
 }
+
+// Execute a pragma for every new sqlite connection
+#[derive(Debug)]
+pub struct PragmaForeignKey;
+impl CustomizeConnection<Connection, ConnError> for PragmaForeignKey {
```
fdb-hiroshima: oh, I was searching for exactly that but did not find it

BaptisteGelez: Looks like it is not working at all :upside_down_face:

fdb-hiroshima: I think your changes are only affecting the web interface, which is at the same time the most important part, and the one not tested by travis :upside_down_face: 
I've tested doing similar changes in plume_model::lib:tests::db (adding `sql_query("PRAGMA foreign_keys = on;").execute(&conn).unwrap();`) which seems to work fine

plume-models/src/blogs.rs
```diff
@@ -829,10 +829,13 @@ pub(crate) mod tests {
                 },
             );
 
+            println!("count bfore {}", blog[0].count_authors(conn));
             user[0].delete(conn, &searcher);
             assert!(Blog::get(conn, blog[0].id).is_some());
             assert!(Blog::get(conn, blog[1].id).is_none());
+            println!("count after bfore {}", blog[0].count_authors(conn));
             user[1].delete(conn, &searcher);
+            println!("count after after {}", blog[0].count_authors(conn));
```
fdb-hiroshima: "bfore", "after bfore", "after after", then is it "after after bfore" or "after after after"?
(joking about the comments, but they should be removed actually, I guess you forgot to)

BaptisteGelez: Oops :grimacing:

PR #351, opened 2018-12-13
Remove useless pagination routes

BaptisteGelez:
Rocket 0.4 let us have routes with optional query parameter

### DIFF
--- a/src/routes/blogs.rs
+++ b/src/routes/blogs.rs
@@ -24,7 +24,8 @@ use template_utils::Ructe;
 use Searcher;
 
 #[get("/~/<name>?<page>", rank = 2)]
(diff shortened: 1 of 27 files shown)


src/routes/blogs.rs
```diff
@@ -118,7 +114,7 @@ pub fn create(conn: DbConn, form: LenientForm<NewBlogForm>, user: User, intl: I1
             is_owner: true
         });
 
-        Ok(Redirect::to(uri!(details: name = slug.clone())))
+        Ok(Redirect::to(uri!(details: name = slug.clone(), page = _)))
```
fdb-hiroshima: does this work to fill an Option with None? I did not know, and was so searching for something like that!

BaptisteGelez: Yes, this new syntax was introduced in the final release of Rocket 0.4 IIRC.

fdb-hiroshima: how ok, missed that

PR #355, opened 2018-12-15
Use SCSS

BaptisteGelez:
- Use variables
- Split everything in various files

The SCSS is compiled with `cargo build`/`run`

I also fixed a few visual issues.


build.rs
```diff
@@ -12,4 +13,12 @@ fn main() {
     println!("cargo:rerun-if-changed=po");
     rocket_i18n::update_po("plume", &["de", "en", "fr", "gl", "it", "ja", "nb", "pl", "ru"]);
     rocket_i18n::compile_po("plume", &["de", "en", "fr", "gl", "it", "ja", "nb", "pl", "ru"]);
+
+    println!("cargo:rerun-if-changed=static/css");
+    let mut out = File::create("static/css/main.css").expect("Couldn't create main.css");
+    println!("annana");
```
fdb-hiroshima: is it some cargo keyword or a debug print?

BaptisteGelez: Oops, I did it again.

It should probably have a pre-commit hook listing the println! I added→ :thinking:

fdb-hiroshima: I personally use "poisson" followed by a color ("poisson rouge" (=golden fish), "poisson vert"...) in my debug prints. That way a simple `git grep poisson` prevent me from doing such. It's quite weird when someone look at my code over my shoulder, but work fairly well

PR #357, opened 2018-12-15
Fix some federation issues

### LINKED ISSUE #172: Following mastodon users won't work
Using my Plume account (fluffy@baptiste.gelez.xyz), I tried to follow my main mastodon account, fluffy@eldritch.cafe, as well as another public account (poutounu@catboy.cafe), and on the Plume interface, it said I followed these accounts, but from mastodon, it doesn't seem so.

I guess there is a federating problem for follows with mastodon. (But a mastodon account can still follow a Plume one.)

### LINKED ISSUE #334: Can't follow plume user from mastodon
It seems like federation with mastodon might be broken again, at least when receiving follow activities

fdb-hiroshima:
There are some bugs in federation, the goal of this will be to fix most of them

(Maybe incomplete) list of things to do : 
 - [x] no notification on follow
 - [x] can't unfollow from Mastodon/Pleroma, but can from Plume
 - [x] Pleroma followers (maybe Mastodon too) are getting notifications from blog post (should only appear in the home timeline)
 - [x] fix #172 
 - [x] fix #334 
 - [x] mention links are a bit broken, they are relative to this instance so displaying them on another instance/ap enabled software result in 404
 - [x] when undoing activity, verify if original activity owner is the undo activity emitter


fdb-hiroshima:
Concerning mention links, I think the only easy solution is to add a domain argument to the parser, and prefix each link with it. Any other solution I'm thinking of either imply post processing or database access in the parser

BaptisteGelez:
Maybe we can always use the FQN for these links so that they are correct on remote instances, and add a redirection when requesting a FQN that is on the current instance?

fdb-hiroshima:
FQN will do the trick when talking to other Plume, but Mastodon use a different model for user links (https://mastodon.local/@<name>, notice no slash between @ and <name>), and Pleroma uses yet another (https://pleroma.local/users/<name>). So either we use users ap-url (need db access), or we use an absolute link to the current instance (not the best solution, but definitely the easiest)

marek-lach:
> FQN will do the trick when talking to other Plume, but Mastodon use a different model for user links (`https://mastodon.local/@<name>`, notice no slash between @ and `<name>`), and Pleroma uses yet another (`https://pleroma.local/users/<name>`). So either we use users ap-url (need db access), or we use an absolute link to the current instance (not the best solution, but definitely the easiest)

Mastodon also recognises the profile URL format of https://mastodon.local/users/<name>, to redirect to the correct user profile, I have tried it. Thus it may be possible if you implement the URL format of https://pleroma.local/users/<name> that it should work just fine for Mastodon too, I think.

fdb-hiroshima:
We could change our scheme, but it wouldn't fix how we treat remote users. `/@/<user@domain>` is valid with Plume, but the equivalent is not with Mastodon/Pleroma. I really think using absolute urls is the easiest way to go

fdb-hiroshima:
One can undo activities made by others. I should fix that too before merging

fdb-hiroshima:
I've no idea why it's that way, I would expect the left panel, /USER/mentions and the notification on /web to be in sync, apparently they aren't???

BaptisteGelez:
Yes, I don't know how Pleroma works :woman_shrugging: but I guess we can ignore this issue, it is still better than before even if not perfect…

PR #363, opened 2018-12-17
Allow for comment deletion

### LINKED ISSUE #224: Delete comments

fdb-hiroshima:
Fix #224 

Current design is a bit terrible. I've tried things, only made it worse. If someone better than me at CSS could give it a try

This tries to conserve as much threads as possible, by re-linking comments to the parent of the one being deleted

### DIFF
--- a/plume-models/src/comments.rs
+++ b/plume-models/src/comments.rs
@@ -285,12 +285,11 @@ impl<'a> Deletable<Connection, Delete> for Comment {
             .expect("Comment::delete: DB error could not update other comments");
         diesel::delete(self)
             .execute(conn)
-            .expect("Post::delete: DB error");
+            .expect("Comment::delete: DB error");
         act
     }
 
     fn delete_id(id: &str, actor_id: &str, conn: &Connection) {
-        println!("Deleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\nDeleting comment\n");
         let actor = User::find_by_ap_url(conn, actor_id);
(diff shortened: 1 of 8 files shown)


igalic:
we could do the reddit thing and leave a greyed out `deleted` box. that might be easiest.

fdb-hiroshima:
My problem is not with deleted comment, I just delete them. It's with the delete button itself

templates/partials/comment.rs.html
```diff
@@ -3,14 +3,14 @@
...
-@(ctx: BaseContext, comm: &Comment, author: User, in_reply_to: Option<&str>)
+@(ctx: BaseContext, comm: &Comment, author: User, in_reply_to: Option<&str>, blog: &str, slug: &str)
 
 <div class="comment u-comment h-cite" id="comment-@comm.id">
     <a class="author u-author h-card" href="@uri!(user::details: name = author.get_fqn(ctx.0))">
         @avatar(ctx.0, &author, Size::Small, true, ctx.1)
         <span class="display-name p-name">@author.name(ctx.0)</span>
         <small>@author.get_fqn(ctx.0)</small>
-    </a>
+	    </a>
```
igalic: why this indent?

plume-models/src/comments.rs
```diff
@@ -1,4 +1,4 @@
-use activitypub::{activity::Create, link, object::Note};
+use activitypub::{activity::{Create, Delete}, link, object::{Note, Tombstone}};
```
igalic: Tombstone… sounds cool

fdb-hiroshima: It's what you need to live behind when deleting things (as opposed to undoing)

templates/partials/comment.rs.html
fdb-hiroshima: There is a 100% chance my vim is still fucking up the indent in html templates, I have no idea why but in those rs.html it absolutely want to add tabs everywhere

fdb-hiroshima:
No, that was all

PR #364, opened 2018-12-18
Comment visibility

### LINKED ISSUE #217: Support for comment visibility
Currently any comment Plume receive is shown on the bottom of the concerned post. Visibility for blog post may be assumed to be public, however comments from other software of the fediverse should probably have their visibility respected

fdb-hiroshima:
Add some support for comment visibility, fix #217 

This add a new column to comment, denoting if they are public or not, and a new table linking private comments to those allowed to read them. There is currently no way to write a private comment from Plume.
Git is having a hard time what happened in Comment::from_activity, but most of it is just re-indentation because a new block was needed to please the borrow checker.
At this point only mentioned users can see private comments, even when posted as "follower only" or equivalent.

What should we do when someone isn't allowed to see a comment? Hide the whole thread, or just the comment? If hiding just the comment, should we mark there is a comment one can't see, but answers they can, or put other comments like if they answered to the same comment the hidden one do?

### DIFF
--- a/migrations/sqlite/2018-12-17-221135_comment_visibility/down.sql
+++ b/migrations/sqlite/2018-12-17-221135_comment_visibility/down.sql
@@ -1,0 +1,18 @@
+-- This file should undo anything in `up.sql`
+CREATE TABLE comments2 (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    content TEXT NOT NULL DEFAULT '',
+    in_response_to_id INTEGER REFERENCES comments(id),
+    post_id INTEGER REFERENCES posts(id) ON DELETE CASCADE NOT NULL,
+    author_id INTEGER REFERENCES users(id) ON DELETE CASCADE NOT NULL,
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url VARCHAR,
+    sensitive BOOLEAN NOT NULL DEFAULT 'f',
+    spoiler_text TEXT NOT NULL DEFAULT ''
+);
+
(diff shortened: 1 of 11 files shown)


migrations/sqlite/2018-12-17-221135_comment_visibility/down.sql
```diff
@@ -0,0 +1,18 @@
...
+    in_response_to_id INTEGER REFERENCES comments(id),
+    post_id INTEGER REFERENCES posts(id) ON DELETE CASCADE NOT NULL,
+    author_id INTEGER REFERENCES users(id) ON DELETE CASCADE NOT NULL,
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url VARCHAR,
+    sensitive BOOLEAN NOT NULL DEFAULT 'f',
+    spoiler_text TEXT NOT NULL DEFAULT ''
+);
+
+INSERT INTO comments2 SELECT id,content,in_response_to_id,post_id,author_id,creation_date,ap_url,sensitive,spoiler_text FROM comments;
```
igalic: if you put these one into each line, my short attention span will be able to follow it 😵

BaptisteGelez:
What you have done seems to work great. :ok_hand: The only little bug that I noticed is that when you are not connected but that there are private comments, the comment area is displayed normally but is empty.

I think that when a comment can not be displayed in a thread, there should a message saying it I think (it would fix the little issue I described too).

fdb-hiroshima:
I've added a message, I'm not too fan of how it looks. I wonder if we shouldn't build the comment "tree"  and trim it according to users rights before rendering, entirely hiding private threads
screenshot_comment_section

BaptisteGelez:
If that's not too complex, yes, it could be a good idea.

PR #365, opened 2018-12-19
Make Plume compile on release

### LINKED ISSUE #328: Plume won't compile on --release
Plume use FromData for String in the two inbox routes, however those are not available in release build as they could be a source of DoS (see [the documentation](https://api.rocket.rs/v0.4/rocket/data/trait.FromData.html#provided-implementations))
Plume version (and/or commit): 74c398d60c9bfeb6799023bb2140c05daf0ae5ff

fdb-hiroshima:
Remove use of String for body parameters
Create SignedJson and implement FromData for it

Fix #328

### DIFF
--- a/plume-common/src/activity_pub/request.rs
+++ b/plume-common/src/activity_pub/request.rs
@@ -41,6 +41,10 @@ impl Digest {
         }
     }
 
+    pub fn verify_header(&self, other: &Digest) -> bool {
+        self.value()==other.value()
+    }
+
     pub fn algorithm(&self) -> &str {
         let pos = self
             .0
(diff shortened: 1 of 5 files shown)


plume-common/src/activity_pub/request.rs
```diff
@@ -41,6 +41,10 @@ impl Digest {
         }
     }
 
+    pub fn verify_header(&self, other: &Digest) -> bool {
+        self.value()==other.value()
+    }
+
     pub fn algorithm(&self) -> &str {
```
igalic: i just realised we have a lot of public functions here, but don't document any

fdb-hiroshima: Yes, we should do it. Also we should probably use https://git.asonix.dog/asonix/http-signatures instead of rolling our own code

fdb-hiroshima:
Maybe we should just test in release, that's what peoples will use anyway

igalic:
except for us developers, so, yeah, 👍

fdb-hiroshima:
Oops I broke code coverage even more I think

marek-lach:
Just posting this here, as it might be relevant for this pull request:

https://mstdn.nipponalba.scot/@jk/101279615979810008

BaptisteGelez:
@marek-lach Thanks! It is indeed related. I responded them.

Issue #296, opened 2018-10-29
Federated Activities should use ID instead of URL


trwnh:
Ideally Mastodon would support Update activities, but there is another solution: federation permalink can be ID-based instead of title-based, and a "pretty url" can be formed from a custom slug containing the title or whatever the user enters.

So, for example, I might post an article on Plume that federates out as `https://plume.example/~/trwnh/6743826481264` and the title is `test post`. The default pretty-url slug would be `test-post` (but I could edit it manually to `test` if I wanted to, or `2018/10/11/1`, or `2018/testpost`, or whatever I wanted) and then the URL would become `https://plume.example/~/trwnh/test-post` (or whatever i set it to)

When federating that out, you would use the `id` as that permalink, and `url` as the slugified link. So for example:
```
"id": "https://plume.example/~/trwnh/6743826481264",
"url": "https://plume.example/~/trwnh/test-post",
```

And the `content` of the `Create` activity would include the `id` instead of the `url`. So something like `<p>$title - $id</p>` at the simplest, and then it would show up in Mastodon like this:

```
trwnh (@trwnh@plume.example) posted 23s ago:
-----------------------------------------------------------------
test post - https://plume.example/~/trwnh/6743826481264
-----------------------------------------------------------------
[reply] [boost] [favorite] [menu]
```

or `trwnh posted a new article:<br>$title<br>$id`, and then it would show up in Mastodon like this:

```
trwnh (@trwnh@plume.example) posted 23s ago:
-------------------------------------------------------
trwnh posted a new article:
test post
https://plume.example/~/trwnh/6743826481264
-------------------------------------------------------
[reply] [boost] [favorite] [menu]
```

The important thing is that the URL that gets federated out cannot change, so it's a bad idea to use a value that can be changed. Not every implementation might support `Update` activities, so using the `id` is more robust and will fail in less cases.

PR #369, opened 2018-12-23
Give reshare and like ap_url before inserting

### LINKED ISSUE #367: Liking and boosting posts does not work
When I like or boost a post (on the same instance), it just takes me to a `domain.com/posturl/{like,boost}` url, which is just a blank page. When I reload the blog's page, the like or boost count still remains at 0, and it seems I haven't affected anything.

fdb-hiroshima:
Fix #367

### DIFF
--- a/plume-models/src/likes.rs
+++ b/plume-models/src/likes.rs
@@ -35,19 +35,6 @@ impl Like {
     find_by!(likes, find_by_ap_url, ap_url as &str);
     find_by!(likes, find_by_user_on_post, user_id as i32, post_id as i32);
 
-    pub fn update_ap_url(&self, conn: &Connection) {
-        if self.ap_url.is_empty() {
-            diesel::update(self)
-                .set(likes::ap_url.eq(format!(
-                    "{}/like/{}",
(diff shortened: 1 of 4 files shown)


fdb-hiroshima:
Yes I did not want to make this too complicated, so I kept older links, but we should move to valid links at some point.

There are two `update_ap_url` left, one is for Post, we need to implement #296 to remove it, as we currently use it on post edition too. The other is for Comment, as we use the comment id in the `ap_url`, we currently need to save a Comment in db before we can know its url. All those should become uuid or similar so we can set them once before insertion, and never edit them again, but I think this is outside of scope for this pr

marek-lach:
A quick question, not sure if related: Does the URL of a post always have to edit itself together with the title of a post? So, when a new post is created, its URL is generated from its first title, right? 

But once the first post URL from that original post title is generated already, then it would help if any subsequent edits to the post title didn't actually change the URL anymore, despite the title being modified.

BaptisteGelez:
I think the URL is only changed if the post is still a draft. If it have been published, a title update shouldn't update the URL.

However, I'm not sure we update the ActivityPub ID when updating the slug/URL, which may cause issues with federation…

PR #370, opened 2018-12-23
An attempt to fix disorted submit button on mobile devices like iPhone - please review

marek-lach:


BaptisteGelez:
@marek-lach Hello! What was the issue you are trying to fix? I can't find anything similar to what you describe in the title…

marek-lach:
**Problem:**
When on a mobile device, like an iPhone, the submit comment button gets disorted (i.e. out of shape, squashed) when you press submit. As far as I know, this isn't the case on the desktop. 

Before pressing the submit button for a comment on an iPhone (though I suspect Android's the same):

img_0551

After pressing the submit button on an iPhone (it works, the distortion is just a visual CSS problem):
img_0552

I am sure the fix for this has to do with responsivness of the submit button, and it must be easy to do, but addmitedly I have no idea, what precisely would fix it, just took a guess.

BaptisteGelez:
This seems to be related to Safari's rendering engine (I cant' reproduce on my Android phone), and I have no iPhone to test. If it doesn't break anything I'll merge your change, hoping it will fix the issue…

marek-lach:
Right, couldn't reproduce the problem on Android. Will try to modify specifically for Safari only then.

marek-lach:
Hmm... reverted the change and tried a Safari-specific style. Let me know if the border is back as before with this last commit.

marek-lach:
I have reverted all of the changes up to now, and added the suggested webkit-specific fix as advised on stack owerflow. As long as this does not disort anything anywhere else, it might work as a quick solution to the problem for now...

PR #374, opened 2018-12-23
Use Rust for the front-end too

BaptisteGelez:
Rust can compile to WASM, so let's use it for front-end code as well.

To compile the front-end:

```
cargo install cargo-web
cd plume-front
cargo web deploy -o ../static
```


static/plume-front.js
```diff
@@ -0,0 +1,572 @@
+"use strict";
```
fdb-hiroshima: I think this file souldn't be committed, it's auto generated and things like `__cargo_web_snippet_199d5eb25dfe761687bcd487578eb7e636bd9650` are probably not reproducible between builds

fdb-hiroshima:
@igalic considering our use-case, probably not much. The main advantage is the whole team being able to debug it I think

BaptisteGelez:
There may be better performances once with the future editor, especially when writing big articles. Another advantage is that we will be able to share code between backend and frontend if we need it.

fdb-hiroshima:
@BaptisteGelez I see Travis failed because of cargo-web installation, you might be interested into this https://github.com/koute/cargo-web#using-cargo-web-on-travis .

.gitignore
```diff
@@ -17,3 +17,6 @@ tags.*
 !tags.rs
 search_index
 main.css
+static/index.html
```
fdb-hiroshima: this line is no longer required

PR #375, opened 2018-12-24
Improve search UI

BaptisteGelez:
- Add a link to the search page in the header
- Translate it

I also removed placeholders because there were not adding much value, and were making the code harder to read when translated (I don't know if it was really a good idea, maybe we should keep them ?).


fdb-hiroshima:
Whoops all this got totally off my mind. I don't like where is placed the search button when connected, but I see no better place to put it. Concerning placeholders, I think it feel a bit empty without them :neutral_face:

BaptisteGelez:
Ok, I'll had placeholders back. And for the button, maybe putting it before the Dashboard button?

BaptisteGelez:
No problem. :slightly_smiling_face:

PR #376, opened 2018-12-24
Change title from Edit account to Create account

marek-lach:


templates/users/new.rs.html
```diff
@@ -15,9 +15,9 @@ <h1>@i18n!(ctx.1, "Create an account")</h1>
             @input!(ctx.1, password (password), "Password", form, errors.clone(), "minlenght=\"8\"")
             @input!(ctx.1, password_confirmation (password), "Password confirmation", form, errors, "minlenght=\"8\"")
 
-            <input type="submit" value="@i18n!(ctx.1, "Create account")" />
+            <input type="submit" value="@i18n!(ctx.1, "Create your account")" />
         </form>
     } else {
-        <p class="center">@i18n!(ctx.1, "Sorry, but registrations are closed on this instance. Try to find another one")</p>
+<p class="center">@i18n!(ctx.1, "Apologies, but registrations are closed on this instance. Visit joinplu.me/#instances, to find a different one")</p>
```
BaptisteGelez: It is a good idea to redirect to joinplu.me, but as it is the link won't work. Could you replace it with something like:

```html
<p class="center">@i18n!(ctx.1, "Apologies, but registrations are closed on this instance. Visit joinplu.me/#instances, to find a different one")</p>
<p class="center">
    <a href="https://joinplu.me/#instances">@i18n!(ctx.1, "Find a different one")</a>
</p>
```

to make it clickable, please? (no problem if you don't, it is still better than before)

marek-lach:
If you sync the English translations from  https://translate.funkwhale.audio/projects/plume/Plume <https://translate.funkwhale.audio/projects/plume/Plume/>
into the master, I’ve already changed and corrected the English translations there. Although I am not sure if you’re using that, so I can of course change the strings on Github directly as well.


24. 12. 2018, 15:32 od notifications@github.com:

BaptisteGelez:
Weblate is not yet synced with this repo, so you'll have to edit it on GitHub for the moment, sorry…

marek-lach:
I have now changed the link to instances to be clickable, while keeping it as a one line. Hopefully it works as intended :-)

templates/users/new.rs.html
```diff
@@ -18,6 +18,7 @@ <h1>@i18n!(ctx.1, "Create an account")</h1>
             <input type="submit" value="@i18n!(ctx.1, "Create your account")" />
         </form>
     } else {
-<p class="center">@i18n!(ctx.1, "Apologies, but registrations are closed on this instance. Visit joinplu.me/#instances, to find a different one")</p>
+<p class="center">@i18n!(ctx.1, "Apologies, but registrations are closed on this instance.
+    <a href="https://joinplu.me/#instances">@i18n!(ctx.1, "Find a different one")</a>")</p>
```
BaptisteGelez: The link should be outside of the `i18n!` call (after the <p>, ideally wrapped in another one to make it centered too), otherwise we will probably end with something really weird (I didn't tested, but it will at least be hard for translators to read, and could result in rendering issue).

templates/users/new.rs.html
```diff
@@ -15,9 +15,10 @@ <h1>@i18n!(ctx.1, "Create an account")</h1>
             @input!(ctx.1, password (password), "Password", form, errors.clone(), "minlenght=\"8\"")
             @input!(ctx.1, password_confirmation (password), "Password confirmation", form, errors, "minlenght=\"8\"")
 
-            <input type="submit" value="@i18n!(ctx.1, "Create account")" />
+            <input type="submit" value="@i18n!(ctx.1, "Create your account")" />
         </form>
     } else {
-        <p class="center">@i18n!(ctx.1, "Sorry, but registrations are closed on this instance. Try to find another one")</p>
+<p class="center">@i18n!(ctx.1, "Apologies, but registrations are closed on this instance.
```
fdb-hiroshima: The closing parenthesis for this i18n call should be at the end of this line, not before the closing `</p>`, nested call won't compile

templates/users/new.rs.html
marek-lach: Ahh, I think I understand now, sorry 👍 
Just pushed another commit to fix this, hopefully a final one for this pull request.

PR #377, opened 2018-12-24
Text strings updated

marek-lach:


marek-lach:
If the build fails now, I'll just remove the clickable link to Plume instances, as it's more trouble, than good...

marek-lach:
Removed the clickable link to Plume's website, as it wasn't able to compile that way, so it's best to leave it.

PR #378, opened 2018-12-24
A small batch of rephrashings for text strings

marek-lach:


marek-lach:
Ah, of course... I'll update texts strings in the tempates too, but they will be created as separate pull requests.

PR #387, opened 2018-12-25
Allow for media cover deletion

### LINKED ISSUE #356: Deleting media partially fail when already used as post cover
Deleting a media that is used as a post cover don't delete the db entry, but delete the file. We should either fail with a message or remove the cover from the post instead.

fdb-hiroshima:
Fix #356

### DIFF
--- a/migrations/sqlite/2018-12-25-164502_media-cover-deletion/down.sql
+++ b/migrations/sqlite/2018-12-25-164502_media-cover-deletion/down.sql
@@ -1,4 +1,21 @@
 -- This file should undo anything in `up.sql`
 
-ALTER TABLE posts DROP CONSTRAINT posts_cover_id_fkey;
-ALTER TABLE posts ADD CONSTRAINT posts_cover_id_fkey FOREIGN KEY (cover_id) REFERENCES medias(id);
+CREATE TABLE posts2 (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    blog_id INTEGER REFERENCES blogs(id) ON DELETE CASCADE NOT NULL,
+    slug VARCHAR NOT NULL,
+    title VARCHAR NOT NULL,
+    content TEXT NOT NULL DEFAULT '',
+    published BOOLEAN NOT NULL DEFAULT 'f',
+    license VARCHAR NOT NULL DEFAULT 'CC-BY-SA',
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url VARCHAR NOT NULL DEFAULT '' UNIQUE,
+    subtitle TEXT NOT NULL DEFAULT '',
+    source TEXT NOT NULL DEFAULT '',
+    cover_id INTEGER REFERENCES medias(id) DEFAULT NULL,
+    CONSTRAINT blog_authors_unique UNIQUE (blog_id, slug)
+);
+
+INSERT INTO posts2 SELECT * from posts;
+DROP TABLE posts;
+ALTER TABLE posts2 RENAME TO posts;
--- a/migrations/sqlite/2018-12-25-164502_media-cover-deletion/down.sql
+++ b/migrations/sqlite/2018-12-25-164502_media-cover-deletion/down.sql
@@ -1,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+
+ALTER TABLE posts DROP CONSTRAINT posts_cover_id_fkey;
+ALTER TABLE posts ADD CONSTRAINT posts_cover_id_fkey FOREIGN KEY (cover_id) REFERENCES medias(id);
--- a/migrations/postgres/2018-12-25-164502_media-cover-deletion/down.sql
+++ b/migrations/postgres/2018-12-25-164502_media-cover-deletion/down.sql
@@ -1,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+
+ALTER TABLE posts DROP CONSTRAINT posts_cover_id_fkey;
+ALTER TABLE posts ADD CONSTRAINT posts_cover_id_fkey FOREIGN KEY (cover_id) REFERENCES medias(id);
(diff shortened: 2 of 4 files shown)


migrations/sqlite/2018-12-25-164502_media-cover-deletion/down.sql
```diff
@@ -0,0 +1,4 @@
+-- This file should undo anything in `up.sql`
+
+ALTER TABLE posts DROP CONSTRAINT posts_cover_id_fkey;
+ALTER TABLE posts ADD CONSTRAINT posts_cover_id_fkey FOREIGN KEY (cover_id) REFERENCES medias(id);
```
igalic: unfortunately, you cannot drop or add constraints on sqlite tables without doing the create temporary table dance

fdb-hiroshima: Travis made me notice it. I knew for columns, I through it worked at least for constraints (it's only drop, add work just fine if I remember well)

BaptisteGelez:
Maybe this change should be integrated in the migrations that introduced constraints, because when you run migrations on an instance with illustrated articles, you get this error:

```
Running migration 2018-12-08-175515_constraints
update or delete on table "medias" violates foreign key constraint "posts_cover_id_fkey" on table "posts"
```

(After seeing this error on baptiste.gelez.xyz I tried to manually fix it, but it just ended in an instance reset :grimacing:)

(I don't think it is a good idea to change previous migrations, but we don't really have choice…)

fdb-hiroshima:
@BaptisteGelez I don't understand, what error did you run into? Is it when an article from before `2018-12-08-175515_constraints` reference a no longer existing media?

BaptisteGelez:
It seem to happen when a media is used both as avatar and post cover. This query deletes it: https://github.com/Plume-org/Plume/blob/master/migrations/postgres/2018-12-08-175515_constraints/up.sql#L51 but it fails because it is used on a post.

EliotBerriot:
I'm facing the same issue described by @BaptisteGelez while applying this migration, and I cannot afford to reset my instance. Is there anything to do to manually fix the issue? I can run SQL by hand if needed :)

BaptisteGelez:
@EliotBerriot try to find user with empty `followers_endpoint` with `SELECT id, username FROM users WHERE followers_endpoint = '';` and if anything is returned, edit the follower endpoint so that it has a value. If I remember correctly, that's how I fixed it on fediverse.blog. Sorry for the bug… :confused:

EliotBerriot:
@BaptisteGelez `SELECT id, username FROM users WHERE followers_endpoint = ''; returned a bunch of users, and I updated it w ith `UPDATE users SET followers_endpoint='noop' WHERE followers_endpoint = '';`

This is the current returns for all the attributes you mentions:

```
plume=# SELECT id, username FROM users WHERE ap_url = '';
 id | username 
----+----------
(0 rows)

plume=# SELECT id, username FROM users WHERE inbox_url = '';
 id | username 
----+----------
(0 rows)

plume=# SELECT id, username FROM users WHERE outbox_url = '';
 id | username 
----+----------
(0 rows)

plume=# SELECT id, username FROM users WHERE followers_endpoint = '';
 id | username 
----+----------
(0 rows)
```

Unfortunately, applying the migrations return the same error again:

```
eliotberriot@peeters:/srv/Plume$ docker-compose run --rm plume diesel migration run --migration-dir migrations/postgres/
Starting plume_postgres_1 ... done
Running migration 2018-12-08-175515_constraints
update or delete on table "medias" violates foreign key constraint "posts_cover_id_fkey" on table "posts"
```

The issues seems to lie within the medias/posts relation, so I'm not sure what it has to do with users?

Anyway, don't be sorry for this: I know the software was still in development and I could face this kind of bug, that was part of the deal :)

BaptisteGelez:
Okay, so maybe try to reset the post covers (it will not impact your articles as you didn't used them for Funkwhale's blog), with `UPDATE posts SET cover_id = null;`. I don't know if it will fix the issue, but it is worth a try.

EliotBerriot:
@BaptisteGelez at least with the `cover_id` thing I get another error: 

```
eliotberriot@peeters:/srv/Plume$ docker-compose run --rm plume diesel migration run --migration-dir migrations/postgres/
Starting plume_postgres_1 ... done
Running migration 2018-12-08-175515_constraints
update or delete on table "comments" violates foreign key constraint "comments_in_response_to_id_fkey" on table "comments"
```

The issues is withing the `comments_in_response_to_id_fkey` constraint now

BaptisteGelez:
I made this script that should delete comments with invalid ActivityPub URLs, and their replies.

```sql
CREATE OR REPLACE FUNCTION delete_comment (id int) RETURNS comments
AS $$
DECLARE
  r int;
BEGIN
  FOR r IN (SELECT id FROM comments WHERE in_response_to_id = id) LOOP
    SELECT delete_comment(r);
  END LOOP;
  DELETE FROM comments WHERE id = id;
END $$ LANGUAGE plpgsql;

SELECT delete_comment(id) FROM comments WHERE ap_url = '';
```

I didn't tested it, but it should only delete broken comments (and hopefully fix the migrations).

EliotBerriot:
@BaptisteGelez unfortunately, this doesn't fix it (for me):

```
eliotberriot@peeters:/srv/Plume$ docker-compose exec postgres psql -U plume
psql (10.5 (Debian 10.5-2.pgdg90+1))
Type "help" for help.

plume=# 
plume=# 
plume=# CREATE OR REPLACE FUNCTION delete_comment (id int) RETURNS comments
plume-# AS $$
plume$# DECLARE
plume$#   r int;
plume$# BEGIN
plume$#   FOR r IN (SELECT id FROM comments WHERE in_response_to_id = id) LOOP
plume$#     SELECT delete_comment(r);
plume$#   END LOOP;
plume$#   DELETE FROM comments WHERE id = id;
plume$# END $$ LANGUAGE plpgsql;
CREATE FUNCTION
plume=# 
plume=# SELECT delete_comment(id) FROM comments WHERE ap_url = '';
 delete_comment 
----------------
(0 rows)

plume=# \q

eliotberriot@peeters:/srv/Plume$ docker-compose run --rm plume diesel migration run --migration-dir migrations/postgres/
Starting plume_postgres_1 ... done
Running migration 2018-12-08-175515_constraints
update or delete on table "comments" violates foreign key constraint "comments_in_response_to_id_fkey" on table "comments"
```

PR #390, opened 2018-12-26
Make Safari more responsive to touch page control

marek-lach:


marek-lach:
I saw that, yes. This just pretty-much hard codes it in, nothing more. As such, it would prevent the scrolling behaviour in mobile Safari to become a momentum based scrolling, i.e. `-webkit-overflow-scrolling: touch` if it was ever to become the default in the future.

That's all it does, it's not important, but it shouldn't really break anything, and it codifies a prefference regardless of what setting may change, or become the default in the future.

You can either merge, or leave it, as you think is best.

PR #391, opened 2018-12-26
Update text strings in tags and admin

marek-lach:


templates/tags/index.rs.html
```diff
@@ -4,8 +4,8 @@
 
 @(ctx: BaseContext, tag: String, articles: Vec<Post>, page: i32, n_pages: i32)
 
-@:base(ctx, i18n!(ctx.1, "Articles tagged \"{0}\""; &tag).as_str(), {}, {}, {
-    <h1>@i18n!(ctx.1, "Articles tagged \"{0}\""; &tag)</h1>
+@:base(ctx, i18n!(ctx.1, "Posts tagged under \"{0}\""; &tag).as_str(), {}, {}, {
```
BaptisteGelez: Is "post" better than "article"? I think we use article most of the time in the user interface, and we should try to be consistent, so if "post" is better it should be used in the rest of the UI as well.

templates/instance/admin.rs.html
```diff
@@ -7,7 +7,7 @@
 
 @(ctx: BaseContext, instance: Instance, form: InstanceSettingsForm, errors: ValidationErrors)
 
-@:base(ctx, i18n!(ctx.1, "Administration of {0}"; instance.name.clone()).as_str(), {}, {}, {
+@:base(ctx, i18n!(ctx.1, "The administration of {0}"; instance.name.clone()).as_str(), {}, {}, {
```
BaptisteGelez: It sounds a bit weird to me, to have "the" in a page title, but maybe that's just me. :woman_shrugging:

PR #392, opened 2018-12-26
Avoid panics

BaptisteGelez:
- Use `Result` as much as possible
- Display errors instead of panicking

TODO (maybe in another PR? this one is already quite big):
- Find a way to merge Ructe/ErrorPage types, so that we can have routes returning `Result<X, ErrorPage>` instead of panicking when we have an `Error`
- Display more details about the error, to make it easier to debug

(sorry, this isn't going to be fun to review, the diff is huge, but it is always the same changes)


fdb-hiroshima:
> (sorry, this isn't going to be fun to review, the diff is huge, but it is always the same changes)

I guess it wasn't fun to write either

plume-models/src/comments.rs
```diff
@@ -53,150 +53,125 @@ impl Comment {
...
         note.object_props
-            .set_to_link_vec(to.clone())
-            .expect("Comment::to_activity: to error");
+            .set_to_link_vec(to.clone())?;
         note.object_props
             .set_tag_link_vec(
                 mentions
                     .into_iter()
-                    .map(|m| Mention::build_activity(conn, &m))
+                    .filter_map(|m| Mention::build_activity(conn, &m).ok())
```
igalic: neat!

plume-models/src/comments.rs
```diff
@@ -216,42 +191,35 @@ impl FromActivity<Note, Connection> for Comment {
...
 
             // save mentions
             if let Some(serde_json::Value::Array(tags)) = note.object_props.tag.clone() {
                 for tag in tags {
                     serde_json::from_value::<link::Mention>(tag)
-                        .map(|m| {
-                            let author = &Post::get(conn, comm.post_id)
-                                .expect("Comment::from_activity: error")
-                                .get_authors(conn)[0];
+                        .map_err(Error::from)
```
igalic: https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err i read the doc, but still don't understand what this does

fdb-hiroshima: I you understand `map`, this is basically the same thing, but for `Err(_)` instead of `Ok(_)`. And passing `Error::from` is the equivalent to `|e| e.from()`

plume-models/src/instance.rs
```diff
@@ -138,7 +130,7 @@ impl Instance {
         open_registrations: bool,
         short_description: SafeString,
         long_description: SafeString,
-    ) {
+    ) -> Result<usize> {
```
fdb-hiroshima: If I'm not mistaken, the usize is going to be how many entries were affected. I'm not sure there is a need to return that?

plume-models/src/instance.rs
```diff
@@ -40,80 +40,72 @@ pub struct NewInstance {
...
-            .id
+            .map_err(Error::from)
     }
 
     insert!(instances, NewInstance);
     get!(instances);
     find_by!(instances, find_by_domain, public_domain as &str);
 
-    pub fn toggle_block(&self, conn: &Connection) {
+    pub fn toggle_block(&self, conn: &Connection) -> Result<usize> {
```
fdb-hiroshima: If I'm not mistaken, the usize is going to be how many entries were affected, which is one?

plume-models/src/instance.rs
fdb-hiroshima: If I'm not mistaken, the usize is going to be how many entries were affected, which is one?

plume-models/src/posts.rs
```diff
@@ -864,113 +839,111 @@ impl Post {
...
                             .object_props
-                            .content_string()
-                            .expect("Post::from_activity: content error"),
+                            .content_string()?,
                     ),
                     published: true,
                     license: license,
                     // FIXME: This is wrong: with this logic, we may use the display URL as the AP ID. We need two different fields
-                    ap_url: article.object_props.url_string().unwrap_or_else(|_|
+                    ap_url: article.object_props.url_string().unwrap_or(
```
fdb-hiroshima: I don't know if there are other places where `unwrap_or_else(|_| ..)` was replaced with `unwrap_or(..)`, if so this comment also apply to them. I think it would be better to `or_else(|_| ..)?`, I think you already did it at other places

plume-models/src/tags.rs
```diff
@@ -27,48 +27,42 @@ impl Tag {
...
-        ht.set_name_string(tag)
-            .expect("Tag::to_activity: name error");
-        ht
+        )))?;
+        ht.set_name_string(tag)?;
+        Ok(ht)
     }
 
-    pub fn delete(&self, conn: &Connection) {
+    pub fn delete(&self, conn: &Connection) -> Result<usize> {
```
fdb-hiroshima: Same about returning Result<()> because it will be 1

plume-models/src/users.rs
```diff
@@ -96,139 +96,109 @@ impl User {
...
+    pub fn one_by_instance(conn: &Connection) -> Result<Vec<User>> {
         users::table
             .filter(users::instance_id.eq_any(users::table.select(users::instance_id).distinct()))
             .load::<User>(conn)
-            .expect("User::one_by_instance: loading error")
+            .map_err(Error::from)
     }
 
-    pub fn delete(&self, conn: &Connection, searcher: &Searcher) {
+    pub fn delete(&self, conn: &Connection, searcher: &Searcher) -> Result<usize> {
```
fdb-hiroshima: same about `Result<()>`

plume-models/src/users.rs
```diff
@@ -96,139 +96,109 @@ impl User {
...
     }
 
-    pub fn get_instance(&self, conn: &Connection) -> Instance {
-        Instance::get(conn, self.instance_id).expect("User::get_instance: instance not found error")
+    pub fn get_instance(&self, conn: &Connection) -> Result<Instance> {
+        Instance::get(conn, self.instance_id)
     }
 
-    pub fn grant_admin_rights(&self, conn: &Connection) {
+    pub fn grant_admin_rights(&self, conn: &Connection) -> Result<usize> {
```
fdb-hiroshima: Same about `Result<()>`

plume-models/src/users.rs
```diff
@@ -96,139 +96,109 @@ impl User {
...
         diesel::update(self)
             .set(users::is_admin.eq(true))
             .execute(conn)
-            .expect("User::grand_admin_rights: update error");
+            .map_err(Error::from)
     }
 
-    pub fn revoke_admin_rights(&self, conn: &Connection) {
+    pub fn revoke_admin_rights(&self, conn: &Connection) -> Result<usize> {
```
fdb-hiroshima: Same about `Result<()>`Same about `Result<()>`

src/main.rs
```diff
@@ -65,10 +68,22 @@ fn init_pool() -> Option<DbPool> {
...
 
 fn main() {
-
     let dbpool = init_pool().expect("main: database pool initialization error");
     let workpool = ScheduledThreadPool::with_name("worker {}", num_cpus::get());
-    let searcher = Arc::new(UnmanagedSearcher::open(&"search_index").unwrap());
+    let searcher = match UnmanagedSearcher::open(&"search_index") {
+        Err(Error::Search(e)) => match e {
+            SearcherError::WriteLockAcquisitionError => panic!(
+r#"Your search index is locked. Plume can't start. To fix this issue, run:
```
fdb-hiroshima: I should have made this message from the start. Maybe add something about making sure no other instance of plume is already running here?

plume-models/src/blogs.rs
```diff
@@ -384,50 +327,38 @@ impl Blog {
...
-    pub fn to_json(&self, conn: &Connection) -> serde_json::Value {
-        let mut json = serde_json::to_value(self).expect("Blog::to_json: serialization error");
-        json["fqn"] = json!(self.get_fqn(conn));
-        json
-    }
-
-    pub fn delete(&self, conn: &Connection, searcher: &Searcher) {
-        for post in Post::get_for_blog(conn, &self) {
-            post.delete(&(conn, searcher));
+    pub fn delete(&self, conn: &Connection, searcher: &Searcher) -> Result<usize> {
```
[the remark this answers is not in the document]
BaptisteGelez: Yes, this result is not used anywhere, it just to have the right return type without having to `.map(|_| ())`, but maybe it's misleading?

plume-models/src/instance.rs
BaptisteGelez: Yes, but https://github.com/Plume-org/Plume/pull/392#discussion_r244369771

plume-models/src/instance.rs
BaptisteGelez: Again, https://github.com/Plume-org/Plume/pull/392#discussion_r244369771

plume-models/src/blogs.rs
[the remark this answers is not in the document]
fdb-hiroshima: I think it is, even more because we did not document any function (nor what they return)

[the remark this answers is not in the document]
BaptisteGelez: Ok, I'll change it then. :+1:

PR #396, opened 2018-12-27
Tags and admin string corrections

marek-lach:
A small, but collected batch of fresh text string updates. Probably the last in such a short period of time :-)

### DIFF
--- a/templates/instance/admin.rs.html
+++ b/templates/instance/admin.rs.html
@@ -25,7 +25,7 @@
       } else {
           <input type="checkbox" name="open_registrations" id="open_registrations">
       }
-      @i18n!(ctx.1, "Allow anyone to register")
+      @i18n!(ctx.1, "Allow anyone to register here")
 
       <label for="short_description">@i18n!(ctx.1, "Short description - byline")<small>@i18n!(ctx.1, "Markdown syntax is supported")</small></label>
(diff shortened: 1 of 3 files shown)


templates/instance/admin.rs.html
```diff
@@ -33,7 +33,7 @@ <h1>@i18n!(ctx.1, "Administration")</h1>
       <label for="long_description">@i18n!(ctx.1, "Long description")<small>@i18n!(ctx.1, "Markdown syntax is supported")</small></label>
       <textarea id="long_description" name="long_description">@Html(form.long_description)</textarea>
 
-      @input!(ctx.1, default_license (text), "Default license", form, errors, "minlenght=\"1\"")
+      @input!(ctx.1, default_license (text), "Default content license", form, errors, "minlenght=\"1\"")
```
BaptisteGelez: The message in the .pot file is "Default article license", you should either add this one to the .pot or change "content" for "article". https://github.com/Plume-org/Plume/blob/master/po/plume.pot#L361

marek-lach:
It seems that 'Allow anyone to register' wasn't even in the .pot file originally? I am not sure where exactly to add it, like at what spot within the .pot file should it be placed? It may be best if you add this phrase into the .pot file to make it translateable then?

Issue #298, opened 2018-10-29
Password reset


BaptisteGelez:
Plume should be able to an email with a temporary password reset link.

https://github.com/mikedilger/mailstrom seems to be a good crate to achieve that.

Plume version (and/or commit): 0.2.0

ShaneKilkelly:
Just a tiny +1 on this, I immediately ran into this issue when signing up for an account.

PR #406, opened 2019-01-02
Add new subcomand for plm to reset password

fdb-hiroshima:
As discussed with @mcrosson on Matrix, while #298 is not fixed, having some way for admin to reset passwords can be usefull

### DIFF
--- a/plume-cli/src/users.rs
+++ b/plume-cli/src/users.rs
@@ -4,6 +4,7 @@ use rpassword;
 use std::io::{self, Write};
 use plume_models::{
     Connection,
+    instance::Instance,
     users::*,
 };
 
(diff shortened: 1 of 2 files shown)


plume-cli/src/users.rs
```diff
@@ -76,3 +91,15 @@ fn new<'a>(args: &ArgMatches<'a>, conn: &Connection) {
...
+
+fn reset_password<'a>(args: &ArgMatches<'a>, conn: &Connection) {
+    let username = args.value_of("name").map(String::from).unwrap_or_else(|| super::ask_for("Username"));
+    let password = args.value_of("password").map(String::from).unwrap_or_else(|| {
+        print!("Password: ");
+        io::stdout().flush().expect("Couldn't flush STDOUT");
+        rpassword::read_password().expect("Couldn't read your password.")
+    });
+    let user = User::find_by_name(conn, &username, Instance::get_local(conn).expect("Failed to get local instance").id)
+        .expect("Failed to get user");
```
igalic: shouldn't we do this bit before asking for a new password? might be a bit frustrating for folks who are prone to typis

PR #417, opened 2019-01-08
Better contrast for input fields

dfeyer:
This change set a white background for input field (text, select and textarea)
to have a better contrast with the page background and improve accessibility for
people with vision problem, and just crappy screen.

This change also remove the round corner for the input to have a clear difference
between editing field and buttons.

This max-width of the form element are a bit bigger too.

### DIFF
--- a/static/css/_forms.scss
+++ b/static/css/_forms.scss
@@ -1,6 +1,6 @@
 label {
  	display: block;
- 	margin: 2em auto 1em;
+ 	margin: 2em auto .5em;
  	font-size: 1.2em;
 }
 input, textarea, select {
--- a/static/css/_forms.scss
+++ b/static/css/_forms.scss
@@ -2,28 +2,24 @@ label {
  	display: block;
  	margin: 2em auto 1em;
  	font-size: 1.2em;
- 	max-width: 40rem;
 }
 input, textarea, select {
  	transition: all 0.1s ease-in;
  	display: block;
  	width: 100%;
- 	max-width: 40rem;
  	margin: auto;
  	padding: 1em;
  	box-sizing: border-box;
 
- 	background: $white;
+ 	background: $form-input-background;
  	color: $black;
  	border: none;
  	border: solid $lightgray thin;
- 	border-radius: 0.5em;
 
  	font-size: 1.2em;
  	font-weight: 400;
 
  	&:focus {
- 	  background: $white;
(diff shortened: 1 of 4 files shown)


dfeyer:
Screenshot are better than a long description

image

static/css/_forms.scss
```diff
@@ -104,6 +100,7 @@ input[type="submit"] { display: block; }
 
 //  Writing page
 form.new-post {
+  max-width: 60em;
```
dfeyer: This will enforce the max-width for all element inside so we don't have to define the max-width for every element that can be used in a form element

static/css/_global.scss
```diff
@@ -47,18 +47,19 @@ small {
 
 /// Main
 body > main > *, .h-feed > * {
- 	padding: 0 20%;
+ 	margin: 0 20%;
```
dfeyer: Needed to set the max-width on form element, should not break anything

static/css/_forms.scss
```diff
@@ -119,7 +116,6 @@ form.new-post {
    	min-height: 20em;
    	overflow-y: hidden;
    	resize: none;
-   	box-sizing: content-box;
```
dfeyer: I don't have the full history of the project so I don't know why you use a different box sizing here ?

dfeyer:
I forgot to say that the margin between to label and the input are smaller with this change, to make the block (label+input/textarea) more "solid".

An other issue we have here is the too low color contrast between the help text and the background, will try to solve this in an other PR

dfeyer:
>  My only concern is that buttons seem a bit inconsistent

@BaptisteGelez  Button are my next topic, will need a better visual language to make them different from input, but not too much. Maybe playing with the height of the element, and use plume violet color for the primary action as a background color. Will push an other PR when my design research are more advanced, will try to prototype on figma and share the design here before the implementation.

BaptisteGelez:
Okay! Then unless you want to fix this little "issue": https://github.com/Plume-org/Plume/pull/417#discussion_r246444928 I will merge this PR (but I'm totally fine merging it as it is if you don't want to take the time to do the change, I completely understand).

dfeyer:
Let's merge this one, and I can fix the margin of cards in an other PR, I hate swapping padding / margin all the time, so maybe we can found a cleaner solution.

PR #418, opened 2019-01-08
Better header line height

dfeyer:
This change also remote all units for line-height to have more consistent styles and better cascading support. This solve the too small line height for multi line header in article and also add styling for h1 to h6 (previously h3 to h6 was not styled at all)

### DIFF
--- a/static/css/_article.scss
+++ b/static/css/_article.scss
@@ -16,7 +16,7 @@ main article {
  	margin: 2.5em auto;
  	font-family: $lora;
  	font-size: 1.2em;
- 	line-height: 1.7em;
+ 	line-height: 1.7;
 
  	a:hover {
  	  text-decoration: underline;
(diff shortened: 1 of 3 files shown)


static/css/_article.scss
```diff
@@ -16,7 +16,7 @@ main article {
  	margin: 2.5em auto;
  	font-family: $lora;
  	font-size: 1.2em;
- 	line-height: 1.7em;
+ 	line-height: 1.7;
```
igalic: what's the measure without a postfix?

dfeyer: line-height without postfix is based on the current font size (so a bit like em), but with the em postfix, cascading is based on the current font size, so the h1-6 get a line height of 1.7em based on the font size of the `main article`selector, so way too small. Without the postfix, element inside the article will have a correct line height, based on their own font size.

Try to insert a really long headline in an article (to have a multi line rendering), with and without this PR to see that the headline are not readable.

See for more details about the unitless value for line-height: https://css-tricks.com/almanac/properties/l/line-height/#article-header-id-0

Issue #229, opened 2018-09-17
Enhancement Proposal: Add Microformats Markup


zcdunn:
> _microformats_ are extensions to HTML for marking up people, organizations, events, locations, blog posts, products, reviews, resumes, recipes etc. Sites use microformats to publish a standard API that is consumed and used by search engines, aggregators, and other tools. 

--[From the indieweb website](https://indieweb.org/microformats)

[Microformats ](http://microformats.org/wiki/microformats2)are simple HTML markup that describe the content. Marking up authors, blogs, and posts would allow interoperability with indieweb site/tools. 

- [ ] [h-card](https://indieweb.org/h-card) mark up for profiles pages and authors in published posts
- [ ] [h-entry](https://indieweb.org/h-entry) mark up for posts, [replies](https://indieweb.org/comments#How_to_markup), etc
- [ ] [in-reply-to](https://indieweb.org/in-reply-to) markup for links from replies to original posts
- [ ] [h-feed](https://indieweb.org/h-feed) mark up for collections of posts

fdb-hiroshima:
I've worked a bit on it, however I'm not familiar with microformat. Could you tell me if what I've done is right? You can test either with the `microformat` branch, or on my [test instance](http://176.143.117.224/) (some urls might be invalid though, because it's usually only running on localhost)

zcdunn:
Hry, @fdb-hiroshima thanks for working on this. I ran a few of the articles on your instance through the parser at http://pin13.net/mf2/. It worked really well, but there were two things I noticed.

* The h-cards didn't have a url item. There should be a link with a 'u-url' class that points to the h-cards subject.
* Every item was listed as having a photo that had the url of the article. That probably means you put 'u-photo' classes on the wrong element. I think the u-photo class should go on the <img> tag

fdb-hiroshima:
So I failed to scope u-photos well, that's why they showed on h-entries instead of h-cards. 
I think I've added the missing u-url where it was missing. Could you also check `/@/Trinity`, `/~/Test` and comments? http://pin13.net/mf2/ helps figuring out some mistakes but as I don't know what I'm supposed to get, I don't know if I'm doing things right

zcdunn:
[...]

PR #419, opened 2019-01-09
Coherent markup for all feeds templates

dfeyer:
Small change to use the same markup for all feeds available in the home page.


dfeyer:
> These elements are here to make it possible for apps based on Microformats to browse Plume articles and feed (see #229). Removing them will break this feature. Maybe adding them to the user's feed could be a solution if you want the markup to be consistent.

And now I remember why h-card was a class name that I know from somewhere else.



> Maybe adding them to the user's feed could be a solution if you want the markup to be consistent.

Make sense, I will update the PR

PR #427, opened 2019-01-12
Fix Plume arm builds

mcrosson:
llvm latest svn commit is broken as of 2019/01/12. This patch pins the llvm+lld versions used for arm compiles.

It also tweaks the architecture detection to be more reliable in multi-lib/multi-arch situations where the CPU is 64bit but the runtime is 32bit.

### DIFF
--- a/script/plume-front.sh
+++ b/script/plume-front.sh
@@ -18,7 +18,7 @@ elif processor == 'armv7l':
 EOF
 )
 
-if [ "$ARCH" == "aarch64" -o "$ARCH" == "armv7l" ] ; then
+if [ $ARCH == "aarch64" -o $ARCH == "armv71" ] ; then
     export PATH=/opt/local/llvm/bin:${PATH}
     cd /app
     RUSTFLAGS="-C linker=lld" cargo web deploy -p plume-front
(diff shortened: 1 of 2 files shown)


script/plume-front.sh
```diff
@@ -1,8 +1,24 @@
...
+if processor == 'aarch64':
+    # Mutli arch arm support is why this 32bit check is present
+    if '32bit' in architecture:
+        print('armv71', end='')
+    else:
+        print('aarch64', end='')
+elif processor == 'x86 64' or processor == 'x86_64':
+    print('amd64', end='')
+elif processor == 'armv7l':
+    print('armhf', end='')
```
igalic: how big is the chance that anyone's still using i386?

mcrosson: @igalic the main linux kernel dev's have suggested dropping i386 support...

If it's something others run into I'm happy to add it back but i need the platform.machine() output to add it to the conditional. I don't have access to i386 and didn't want to add an arch that I don't have valid output for.

script/plume-front.sh
```diff
@@ -1,8 +1,24 @@
 #!/bin/bash
 
-ARCH=`arch`
+ARCH=$(python <<EOF
```
BaptisteGelez: Wouldn't it be possible to do the same with Bash? And maybe you could use Python 3 if it present instead of importing from `__future__`.

igalic: > Wouldn't it be possible to do the same with Bash?

yes. with great difficulty.

> And maybe you could use Python

+1

PR #430, opened 2019-01-18
Better big form handling

### LINKED ISSUE #425: "422: Unprocessable Entity" - inssuficient error message
I'm trying to test some markdown. Primarily I used [this document](http://daringfireball.net/projects/markdown/syntax.text)
from https://stackoverflow.com/questions/2238012/

However upon pressing submit I'm redirected to 422 error message that doesn't really give any real information:

> 422: Unprocessable Entity
> The request was well-formed but was unable to be followed due to semantic errors.

I've deleted around the page source and something is breaking the post; would be nice if the error message could say what is breaking the submission.  

Same with commonmark's spec file: https://raw.githubusercontent.com/commonmark/CommonMark/master/spec.txt

fdb-hiroshima:
Allow customizing max form size from env vars

fix #425 
I should also add a proper error page before merging

### DIFF
--- a/plume-front/src/main.rs
+++ b/plume-front/src/main.rs
@@ -1,16 +1,17 @@
+#![recursion_limit="128"]
 #[macro_use]
 extern crate stdweb;
 
-use stdweb::{unstable::TryFrom, web::{*, event::*}};
+use stdweb::{unstable::{TryFrom, TryInto}, web::{*, event::*}};
 
 fn main() {
-    auto_expand();
+    editor_loop();
     menu();
     search();
 }
 
-/// Auto expands the editor when adding text
-fn auto_expand() {
+/// Auto expands the editor when adding text and count chars
+fn editor_loop() {
     match document().query_selector("#plume-editor") {
         Ok(Some(x)) => HtmlElement::try_from(x).map(|article_content| {
             let offset = article_content.offset_height() - (article_content.get_bounding_client_rect().get_height() as i32);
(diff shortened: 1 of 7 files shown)


igalic:
aaaaand perhaps a js counter?

src/main.rs
```diff
@@ -95,7 +99,17 @@ Then try to restart Plume.
         exit(0);
     }).expect("Error setting Ctrl-c handler");
 
-    rocket::ignite()
+    let mut config = Config::active().unwrap();
+    config.set_address(env::var("ROCKET_ADDRESS").unwrap_or_else(|_| "localhost".to_owned())).unwrap();
+    config.set_port(env::var("ROCKET_PORT").ok().map(|s| s.parse::<u16>().unwrap()).unwrap_or(8000));
```
BaptisteGelez: Not really related, but we are saying in the docs it should be `7878` by default (even if it is not currently the case :grimacing:), maybe it is a good occasion to change it?

fdb-hiroshima: it is indeed. Also do you know of other rocket env vars I might have forgotten?

fdb-hiroshima:
I've added a char counter, however I had no idea where to put it, it's currently on the top of the textarea, thinking about it that's maybe the worst place to put it after
For some reason Rocket seems to accept a few bytes more than configured, which I don't take into account, so it's possible to push an article with -1 or -2 chars left

plume-front/src/main.rs
```diff
@@ -1,16 +1,17 @@
+#![recursion_limit="128"]
```
igalic: what's this used for?

plume-front/src/main.rs
```diff
@@ -19,7 +20,33 @@ fn auto_expand() {
                 js! {
                     @{&article_content}.style.height = "auto";
                     @{&article_content}.style.height = @{&article_content}.scrollHeight - @{offset} + "px";
-                }
+                };
+                window().set_timeout(|| {match document().query_selector("#post-form") {
+                    Ok(Some(form)) => HtmlElement::try_from(form).map(|form| {
+                        if let Some(len) = form.get_attribute("content-size").and_then(|s| s.parse::<i32>().ok()) {
+                            let consumed: i32 = js!{
```
igalic: whoa! `js!`?!

plume-front/src/main.rs
BaptisteGelez: Macros can be recursive in Rust, but to make sure every build has an end, the compiler defines a macro recursion limit. This line just tells Rust to make it a bit higher, because the `js!` macro calls itself a lot of time.

plume-front/src/main.rs
BaptisteGelez: Yes, it is macro to call JS code in WASM, when you are too limited by Rust APIs (since they are still quite experimental).

fdb-hiroshima: the rational here is that stdweb doesn't support [HTMLFormElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement) yet, and querying recursively every child to test if it's an input is annoying. And to not pass variables back and fourth a lot of time, I made most computations in js, those are only a few add anyway.
If you want to read it, things in `@{}` are rust vars (and possibly code I think), other things are plain js

PR #431, opened 2019-01-20
Update rocket_i18n and add gettext_macros

BaptisteGelez:
Internationalization now uses proc-macros that generate the .pot file
automatically.


fdb-hiroshima:
I'm getting 
```rust
GET / text/html:
    => Matched: GET / (index)
    => Error: Attempted to retrieve unmanaged state!
thread '<unnamed>' panicked at 'Outcome::expect() failed: Couldn't retrieve translations because they are not managed by Rocket.', /home/trinity/.cargo/registry/src/github.com-1ecc6299db9ec823/rocket-0.4.0/src/outcome.rs:185:18
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```
on every page. Any idea on what I did wrong? (or does it not work??)

fdb-hiroshima:
`include_i18n!` returns a `Translation` according to docs, but routes try to load an `I18n`, that may be the issue, but I don't know how you can not get the error

BaptisteGelez:
I can indeed reproduce. I'll try to fix the issue.

BaptisteGelez:
Ok, so the issue was just that I tested with my local version of gettext-macros, but there were a few changes that were not yet on crates.io.

I'll try to fix the conflict now…

.buildconfig
```diff
@@ -0,0 +1,10 @@
+[default]
+name=Default
+runtime=host
+config-opts=
+run-opts=
+prefix=/home/elza/.cache/gnome-builder/install/plume/host
```
fdb-hiroshima: I don't know what buildconfig is for, but it seems quiet tied to your specific configuration, can't it be made otherwise?

PR #443, opened 2019-02-11
Big refactoring of the Inbox

BaptisteGelez:
We now have a type that routes an activity through the registered handlers
until one of them matches.

Each Actor/Activity/Object combination is represented by an implementation of AsObject

These combinations are then registered on the Inbox type, which will try to deserialize
the incoming activity in the requested types.

Advantages:
- nicer syntax: the final API is clearer and more idiomatic
- more generic: only two traits (`AsActor` and `AsObject`) instead of one for each kind of activity
- we always try to dereference unknown objects
- it is easier to see which activities we handle and which one we don't

(sorry for the big diff once again :confused:)


plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,439 @@
...
+///
+/// # Panics
+///
+/// This function panics if the value is neither a string nor an object with an
+/// `id` field that is a string.
+fn get_id<'a>(json: serde_json::Value) -> String {
+    match json {
+        serde_json::Value::String(s) => s,
+        serde_json::Value::Object(map) => get_id(map["id"].clone()),
+        // I don't know if it is really a good idea to panic here…
```
fdb-hiroshima: as it is more of a lib kind, I don't think it is a good idea. Returning an Option<String> and `.expect()`ing on it would be better if possible

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,439 @@
...
+/// Otherwise we return its `id` field, if it is a string
+///
+/// # Panics
+///
+/// This function panics if the value is neither a string nor an object with an
+/// `id` field that is a string.
+fn get_id<'a>(json: serde_json::Value) -> String {
+    match json {
+        serde_json::Value::String(s) => s,
+        serde_json::Value::Object(map) => get_id(map["id"].clone()),
```
fdb-hiroshima: is it calling itself recursively? a `{"id":{"id":{"id":{"id":"something"}}}` would be considered valid and having an id if so, and as it is cloning, with a big enough json, memory usage could grow quadratic to the size of the original payload

plume-models/src/blogs.rs
```diff
@@ -384,7 +387,7 @@ impl WithInbox for Blog {
     }
 
     fn is_local(&self) -> bool {
-        self.instance_id == 0
+        self.instance_id == 1 // TODO: this is not always true
```
fdb-hiroshima: why was it 0 before and 1 now?

plume-models/src/lib.rs
```diff
@@ -140,10 +141,37 @@ impl From<std::io::Error> for Error {
     }
 }
 
+impl From<plume_common::activity_pub::inbox::InboxError> for Error {
```
fdb-hiroshima: this could use a `use` statement :thinking:

plume-models/src/lib.rs
```diff
@@ -306,7 +334,9 @@ pub fn ap_url(url: &str) -> String {
 #[cfg(test)]
 #[macro_use]
 mod tests {
-    use diesel::{dsl::sql_query, Connection, RunQueryDsl};
+    #[cfg(feature = "sqlite")]
```
fdb-hiroshima: does this mean no more warnings when building with pg backend?

plume-models/src/mentions.rs
```diff
@@ -3,7 +3,7 @@ use diesel::{self, ExpressionMethods, QueryDsl, RunQueryDsl};
 
 use comments::Comment;
 use notifications::*;
-use plume_common::activity_pub::inbox::Notify;
+// use plume_common::activity_pub::inbox::Notify;
```
fdb-hiroshima: ?

plume-models/src/blogs.rs
BaptisteGelez: This function is never used I think, so we didn't noticed this bug before, but the first ID in databases is 1, never 0.

plume-models/src/lib.rs
BaptisteGelez: Normally, yes. :slightly_smiling_face:

plume-models/src/mentions.rs
BaptisteGelez: I guess I temporally "disabled" it to see what errors I got or something like that, and I forgot to completely remove it. :grimacing: (I should probably review my own code before pushing it)

src/routes/comments.rs
```diff
@@ -101,11 +105,15 @@ pub fn create(blog_name: String, slug: String, form: LenientForm<NewCommentForm>
...
 
 #[post("/~/<blog>/<slug>/comment/<id>/delete")]
-pub fn delete(blog: String, slug: String, id: i32, user: User, conn: DbConn, worker: Worker) -> Result<Redirect, ErrorPage> {
+pub fn delete(blog: String, slug: String, id: i32, user: User, conn: DbConn, worker: Worker, searcher: Searcher) -> Result<Redirect, ErrorPage> {
     if let Ok(comment) = Comment::get(&*conn, id) {
         if comment.author_id == user.id {
             let dest = User::one_by_instance(&*conn)?;
-            let delete_activity = comment.delete(&*conn)?;
+            let delete_activity = comment.build_delete(&*conn)?;
+            Inbox::handle(&Context::build(&conn, &searcher), serde_json::to_value(&delete_activity).map_err(Error::from)?)
```
fdb-hiroshima: this line seems to be failing on my local instance. I haven't searched why yet

src/inbox.rs
```diff
@@ -24,140 +13,110 @@ use serde_json;
...
+    Post => Post,
+    Reshare => Reshared
+}
+
+fn inbox(conn: &Connection, search: &Searcher, act: serde_json::Value) -> Result<InboxResult, Error> {
+    Inbox::handle(&Context::build(conn, search), act)
+        .with::<User, Announce, Post,            _>()
+        .with::<User, Create,   Comment,         _>()
+        .with::<User, Create,   Post,            _>()
+        .with::<User, Delete,   Comment,         _>()
```
fdb-hiroshima: Deleting comment seems to return a `Shared inbox error: Inbox(InvalidObject)`

src/routes/posts.rs
```diff
@@ -407,7 +410,11 @@ pub fn delete(blog_name: String, slug: String, conn: DbConn, user: User, worker:
             Ok(Redirect::to(uri!(details: blog = blog_name.clone(), slug = slug.clone(), responding_to = _)))
         } else {
             let dest = User::one_by_instance(&*conn)?;
-            let delete_activity = post.delete(&(&conn, &searcher))?;
+            let delete_activity = post.build_delete(&*conn)?;
+            Inbox::handle(&Context::build(&*conn, &searcher), serde_json::to_value(&delete_activity).map_err(Error::from)?)
```
fdb-hiroshima: This seems to be failing too

src/inbox.rs
```diff
@@ -24,140 +13,110 @@ use serde_json;
...
+        .with::<User, Delete,   Post,            _>()
+        .with::<User, Follow,   User,            _>()
```
fdb-hiroshima: this doesn't seems to work

src/inbox.rs
```diff
@@ -24,140 +13,110 @@ use serde_json;
...
+        .with::<User, Like,     Post,            _>()
```
fdb-hiroshima: this doesn't seems to work

src/inbox.rs
```diff
@@ -24,140 +13,110 @@ use serde_json;
...
+    Reshare => Reshared
+}
+
+fn inbox(conn: &Connection, search: &Searcher, act: serde_json::Value) -> Result<InboxResult, Error> {
+    Inbox::handle(&Context::build(conn, search), act)
+        .with::<User, Announce, Post,            _>()
+        .with::<User, Create,   Comment,         _>()
+        .with::<User, Create,   Post,            _>()
+        .with::<User, Delete,   Comment,         _>()
+        .with::<User, Delete,   Post,            _>()
```
fdb-hiroshima: Deleting post seems to return a `Shared inbox error: Inbox(InvalidObject)`

src/inbox.rs
```diff
@@ -24,140 +13,110 @@ use serde_json;
...
+    Comment => Commented,
+    follows::Follow => Followed,
+    likes::Like => Liked,
+    Post => Post,
+    Reshare => Reshared
+}
+
+fn inbox(conn: &Connection, search: &Searcher, act: serde_json::Value) -> Result<InboxResult, Error> {
+    Inbox::handle(&Context::build(conn, search), act)
+        .with::<User, Announce, Post,            _>()
```
fdb-hiroshima: this doesn't seems to work

src/routes/comments.rs
BaptisteGelez: If it is exactly on this line, the only thing that can fail is the JSON serialization I think. But that doesn't make sense?? I'll see if I can reproduce.

BaptisteGelez:
Thanks for your review @fdb-hiroshima. I'll try to reproduce your issues and fix them, and maybe add tests for it in plume-models too.

src/routes/comments.rs
fdb-hiroshima: I've added a debug print on the line above and the line under, I get one but not the other (I consider with::<...> and done to be on the same line, they are the same statement)

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,560 @@
...
 
-pub trait Notify<C> {
-    type Error;
+/*
+ Type arguments:
+ - C: Context
+ - E: Error
+ - R: Result
+*/
+impl<'a, C, E, R> Inbox<'a, C, E, R> where E: From<InboxError<E>> + Debug {
```
igalic: what does the `where` clause here mean? I've never seen `+ Debug` @_@

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,560 @@
...
+///
+/// - `C`: the context to be passed to this activity handler from the `Inbox` (usually a database connection)
+pub trait AsActor<C>: Sized {
+    /// Return the URL of this actor's inbox
     fn get_inbox_url(&self) -> String;
 
-    fn get_shared_inbox_url(&self) -> Option<String>;
+    /// If this actor has shared inbox, its URL should be returned by this function
+    fn get_shared_inbox_url(&self) -> Option<String> {
+        None
```
igalic: is this missing an implementation, or is this it? 😜

plume-models/src/blogs.rs
igalic: Instance should probably have a method that returns the local `instance_id`

plume-common/src/activity_pub/inbox.rs
fdb-hiroshima: This mean that E must Implement `From<InboxError<E>>` and `Debug` (`Debug` allow ugly display via `println!("{:?}", yourStruct)`) for the functions in this impl block to be available

plume-common/src/activity_pub/inbox.rs
BaptisteGelez: No it is the default implementation: unless the implementation specifies a shared inbox URL, it is `None` (but in reality I think all our implementations override this behavior)

BaptisteGelez:
I fixed all the conflicts, and now instead of using `Context` we use `PlumeRockets`. I also deployed this branch to https://baptiste.gelez.xyz to make testing easier if needed.

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,582 @@
...
+
+    /// Registers an handler on this Inbox.
+    pub fn with<A, V, M>(self) -> Inbox<'a, C, E, R>
+    where
+        A: AsActor<&'a C> + FromId<C, Error = E>,
+        V: activitypub::Activity,
+        M: AsObject<A, V, &'a C, Error = E> + FromId<C, Error = E>,
+        M::Output: Into<R>,
+    {
+        if let Inbox::NotHandled(ctx, act, e) = self {
```
igalic: is there such a thing as `if not let`? so we can move this whole block out by one indent, by returning early?

plume-models/src/comments.rs
```diff
@@ -265,25 +310,40 @@ impl FromActivity<Note, Connection> for Comment {
         }
 
         comm.notify(conn)?;
+        dbg!("hey 6");
```
igalic: debug?

plume-models/src/comments.rs
```diff
@@ -265,25 +310,40 @@ impl FromActivity<Note, Connection> for Comment {
...
+                m.delete(conn)?;
+            }
+            diesel::update(comments::table)
+                .filter(comments::in_response_to_id.eq(self.id))
+                .set(comments::in_response_to_id.eq(self.in_response_to_id))
+                .execute(conn)?;
+            diesel::delete(&self).execute(conn)?;
+            Ok(())
+        } else {
+            Err(Error::Unauthorized)
```
igalic: if author != reiten Err

and remove the entire of indent, pls

plume-common/src/activity_pub/inbox.rs
BaptisteGelez: I don't think so…

plume-models/src/comments.rs
BaptisteGelez: Oops.

plume-common/src/activity_pub/inbox.rs
fdb-hiroshima: There is match

BaptisteGelez:
So I did some testing, and here are the results:

|                | Mastodon |                 Pleroma                 |      Plume     |
|:--------------:|:--------:|:---------------------------------------:|:--------------:|
|    Announce    |     ✔️    |                    ✔️                    |        ✔️       |
|   Create Note  |   ✔️ (1)  |                  ✔️ (1)                  |      ✔️ (1)     |
| Create Article |     -    |                    -                    | `InvalidObject`|
|   Delete Note  |   ✔️ (2)  |                  ✔️ (2)                  |      ✔️ (2)     |
| Delete Article |     -    |                    -                    | Could not test |
|     Follow     |   ✔️ (3)  |                    ✔️                    |        ✔️       |
|      Like      |     ✔️    |                    ✔️                    |        ✔️       |
|  Undo Announce |     ✔️    |              `InvalidObject`            |  `InvalidObject`|
|   Undo Follow  |     ✔️    |                    ✔️                    |        ✔️       |
|    Undo Like   |     ✔️    |  No error in the logs, but doesn't work |        ✔️       |
| Update Article |     -    |                    -                    |  Couldn't test |

1. The CW is not here
2. The notification is not deleted (see #499)
3. There is no error, but after refreshing the Mastodon page, the follow is not accepted (:hourglass: icon)

I will try to debug a bit more the errors I encountered.

igalic:
it would be cool if we could add a `service: mastodon` and `service: pleroma` to Travis 😅

BaptisteGelez:
I think if we were on Gitlab we could just pull their Docker images… I really miss Gitlab CI.

fdb-hiroshima:
can't we do it the ugly way with some bash script?

BaptisteGelez:
We could, but it would take a lot of time to install all the dependencies, and the CI already takes so long to run.

igalic:
i think we're starting to discuss integration tests, and it's not really fruitful in this PR.

plume-models/src/users.rs
```diff
@@ -731,23 +731,17 @@ impl FromId<PlumeRocket> for User {
...
 
-        if acct
-            .object
-            .ap_actor_props
-            .preferred_username_string()?
-            .contains(&['<', '>', '&', '@', '\'', '"'][..])
-        {
+        let username = acct.object.ap_actor_props.preferred_username_string()?;
+
+        if username.contains(&['<', '>', '&', '@', '\'', '"'][..]) {
```
igalic: and space.

BaptisteGelez:
So, I fixed most issues (and updated the table above). I just need to understand why Pleroma is failing to undo like/boost, fix the space issue @igalic found (even if it is not directly related to this PR, it was here before), and do some optimizations (the current algorithm can dereference the same URL multiple times while handling the same activity for the moment). After that, I think this PR will be ready for real.

BaptisteGelez:
Update: I optimized the algorithm a little bit to avoid dereferencing the same object twice, fixed the space-in-username issue, and found why it was not possible to Undo Announce and Like from Pleroma. It was just not sending the activity to Plume because no one was following the Pleroma account on the Plume instance. It works fine if I follow my Pleroma account from Plume. I still created a [bug report](https://git.pleroma.social/pleroma/pleroma/issues/779).

plume-models/src/lib.rs
```diff
@@ -344,3 +378,4 @@ pub mod schema;
 pub mod search;
 pub mod tags;
 pub mod users;
+pub use plume_rocket::PlumeRocket;
```
igalic: fmt will not be happy with this :P

BaptisteGelez: Yes, I saw that. Fixed it. ^^

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,609 @@
...
 }
 
-pub trait Deletable<C, A> {
-    type Error;
+/// Get the ActivityPub ID of a JSON value.
+///
+/// If the value is a string, its value is returned.
+/// Otherwise we return its `id` field, if it is a string
+///
+/// # Panics
```
fdb-hiroshima: I don't think it can actually panic? Also the `Otherwise we return its 'id' field, if it is a string` is a bit unclear

plume-common/src/activity_pub/inbox.rs
```diff
@@ -1,48 +1,609 @@
...
+///
+/// The two functions to implement are `from_activity` to create (and save) a new object
+/// of this type from its AP representation, and `from_db` to try to find it in the database
+/// using its ID.
+///
+/// When dealing with the "object" field of incoming activities, `Inbox` will try to see if it is
+/// a full object, and if so, save it with `from_activity`. If it is only an ID, it will try to find
+/// it in the database with `from_db`, and otherwise dereference (fetch) the full object and parse it
+/// with `from_activity`.
+pub trait FromId<C>: Sized {
```
fdb-hiroshima: Isn't Sized implied by default?

plume-common/src/activity_pub/inbox.rs
BaptisteGelez: Yeah, I think I forgot to update the docs.

plume-common/src/activity_pub/inbox.rs
BaptisteGelez: IIRC, the compiler threw errors when it wasn't here. :woman_shrugging:

BaptisteGelez: I'll try to remove it just to be sure.

BaptisteGelez: It was explicitly needed here, to be able to have `Result<Self, _>` here, but was not actually required on the two other types.

fdb-hiroshima: ok, I did not know

plume-models/src/blogs.rs
```diff
@@ -409,7 +298,106 @@ impl IntoId for Blog {
...
+    type Object = CustomGroup;
+
+    fn from_db(c: &PlumeRocket, id: &str) -> Result<Self> {
+        Self::find_by_ap_url(&c.conn, id)
+    }
+
+    fn from_activity(c: &PlumeRocket, acct: CustomGroup) -> Result<Self> {
+        let url = Url::parse(&acct.object.object_props.id_string()?)?;
+        let inst = url.host_str()?;
+        let instance = Instance::find_by_domain(&c.conn, inst).or_else(|_| {
```
fdb-hiroshima: out of scope for this pr, but we should consider a `find_or_create` of some kind instead of `find_by_domain.or_else(insert)`, so to not tie to much code from one part of the model to another

plume-models/src/blogs.rs
```diff
@@ -703,7 +693,9 @@ pub(crate) mod tests {
             )
             .unwrap();
 
-            assert_eq!(Blog::find_by_fqn(conn, "SomeName").unwrap().id, blog.id);
+            dbg!(blog.fqn);
```
fdb-hiroshima: is this wanted or leftover?

plume-models/src/posts.rs
```diff
@@ -131,13 +128,15 @@ impl<'a> Provider<(&'a Connection, &'a Worker, &'a Searcher, Option<i32>)> for P
...
+            .get_results::<Post>(conn)
             .map(|ps| {
                 ps.into_iter()
                     .filter(|p| {
                         p.published
-                            || user_id
-                                .map(|u| p.is_author(conn, u).unwrap_or(false))
+                            || rockets
+                                .user
+                                .clone()
```
fdb-hiroshima: I think we should try to chase clones everywhere, a lot of theme could be as_ref, which more efficient in term of both memory and computation

plume-models/src/posts.rs
```diff
@@ -131,13 +128,15 @@ impl<'a> Provider<(&'a Connection, &'a Worker, &'a Searcher, Option<i32>)> for P
...
+                                .map(|u| p.is_author(conn, u.id).unwrap_or(false))
```
fdb-hiroshima: wouldn't a `and_then` be more appropriate?

plume-models/src/posts.rs
```diff
@@ -68,17 +67,17 @@ pub struct NewPost {
...
+    fn get(rockets: &PlumeRocket, id: i32) -> ApiResult<PostEndpoint> {
+        let conn = &*rockets.conn;
         if let Ok(post) = Post::get(conn, id) {
             if !post.published
-                && !user_id
-                    .map(|u| post.is_author(conn, u).unwrap_or(false))
+                && !rockets
+                    .user
+                    .clone()
+                    .map(|u| post.is_author(conn, u.id).unwrap_or(false))
```
fdb-hiroshima: wouldn't a and_then be more appropriate?

plume-models/src/posts.rs
```diff
@@ -925,154 +873,365 @@ impl Post {
...
+        let post = Post::insert(
+            conn,
+            NewPost {
+                blog_id: blog?.id,
+                slug: title.to_kebab_case(),
+                title,
+                content: SafeString::new(&article.object_props.content_string()?),
+                published: true,
+                license,
+                // FIXME: This is wrong: with this logic, we may use the display URL as the AP ID. We need two different fields
```
fdb-hiroshima: is it related to #296 ?

po/plume-front/en.po~
```diff
@@ -0,0 +1,45 @@
+msgid ""
```
fdb-hiroshima: aren't ~ files backup some editor make?

plume-models/src/blogs.rs
BaptisteGelez: leftover…

plume-models/src/posts.rs
BaptisteGelez: More or less I think. Adding a `url` field to posts could solve both issues.

po/plume-front/en.po~
BaptisteGelez: These files are generated by the gettext command line tools, but they are normally gitignored. I'll remove them.

fdb-hiroshima:
### Federation with old Plume
#### Sending activities
 - Unsubscribing from a user does not seem to delete the entry in database
 - Sending comment don't seem to propagate CW (maybe bug from old Plume)
#### Receiving
all right

### Federation with itself
 - Unsubscribing from a user does not seem to delete the entry in database
 - Sending/Receiving Post Update does not seems to update post cover-picture

### pleroma
todo

plume-models/src/posts.rs
```diff
@@ -208,15 +208,10 @@ impl Provider<PlumeRocket> for Post {
         let domain = &Instance::get_local(&conn)
             .map_err(|_| ApiError::NotFound("posts::update: Error getting local instance".into()))?
             .public_domain;
-        let author = User::get(
```
fdb-hiroshima: wow I missed how ugly this thing was

fdb-hiroshima:
https://github.com/Plume-org/Plume/pull/443#issuecomment-476341573 by the way if we manage to get some certificate win can use in tests, this is now probably possible with circleci

BaptisteGelez:
@fdb-hiroshima The issues with Pleroma are normally fixed now.

BaptisteGelez:
Yay, Finally! Thank you @igalic and @fdb-hiroshima for reviewing this big PR! :blush:

PR #444, opened 2019-02-11
Add a page listing people someone follows

### LINKED ISSUE #325: Add a page listing followed users
We currently only have followers.

BaptisteGelez:
Nothing exceptional, the layout is the same as the followers page.

Fixes #325


plume-models/src/users.rs
```diff
@@ -577,6 +577,28 @@ impl User {
             .map_err(Error::from)
     }
 
+    pub fn count_followed(&self, conn: &Connection) -> Result<i64> {
```
igalic: n.b.: many people on the fediverse are not happy with the "clout" that comes from these numbers

do we offer an option to disable it?

i would also suggest considering "subscribe" instead of "follow".
it fits better with a blog, and has slightly less of a cult meaning

BaptisteGelez: These numbers are not currently displayed in the UI, we only use this function to know how many pages listing followers/followed we need. If we display them one day, I totally agree that we should be able to hide them (maybe even hide them by default?), or that they should only be shown for your own profile.

And I agree on "follow"/"subscribe" too, I will change it. :slightly_smiling_face:

po/plume/de.po
```diff
@@ -315,12 +315,12 @@ msgstr ""
...
-msgid "Followers"
-msgstr "Follower"
+msgid "Subscribers"
+msgstr ""
 
 #, fuzzy
-msgid "Followed"
-msgstr "Folgen"
+msgid "Subscriptions"
+msgstr "Beschreibung"
```
igalic: not quite… "Abonoment" would be a better translation.

po/plume/de.po
```diff
@@ -315,12 +315,12 @@ msgstr ""
 msgid "Articles"
 msgstr "Artikel"
 
-msgid "Followers"
-msgstr "Follower"
+msgid "Subscribers"
+msgstr ""
```
igalic: "Abonomenten"

po/plume/de.po
```diff
@@ -342,12 +342,11 @@ msgstr "Ändere dein Profil"
...
-#, fuzzy
-msgid "Unfollow"
-msgstr "Folgen"
+msgid "Unsubscribe"
+msgstr ""
 
-msgid "Follow"
-msgstr "Folgen"
+msgid "Subscribe"
+msgstr ""
```
igalic: "abonieren"

po/plume/de.po
```diff
@@ -342,12 +342,11 @@ msgstr "Ändere dein Profil"
 msgid "Open on {0}"
 msgstr ""
 
-#, fuzzy
-msgid "Unfollow"
-msgstr "Folgen"
+msgid "Unsubscribe"
+msgstr ""
```
igalic: "abbestellen"

po/plume/de.po
BaptisteGelez: Yes, these strings have probably been automatically translated by GetText… :confused: I will edit them.

BaptisteGelez:
I replaced the other occurrences of "follow" while we are at it.

PR #446, opened 2019-02-15
Update nodeinfo

### LINKED ISSUE #433: Nodeinfo improvements and fix
Plume version (and/or commit): latest

Fix:
- Nodeinfo `software.name` should be lowercase
- `openRegistrations` seems to always be true even if deactivated

Improvements:
Add some metadata:
- `nodeName`: instance name
- `nodeDescription`: instance short description

zcdunn:
Fix #433

I added the repo link to Cargo.toml so that `software.repository` could be configurable like @rhaamo suggested. I don't know if it's ok to include `software.repository` without bumping the schema version, but I didn't know if that would break any clients that parse nodeinfo with a hardcoded schema version.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,6 +2,7 @@
 authors = ["Plume contributors"]
 name = "plume"
 version = "0.2.0"
+repository = "https://github.com/Plume-org/Plume"
 
 [dependencies]
 activitypub = "0.1.3"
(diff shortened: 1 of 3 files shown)


rhaamo:
`software.repository` is nodeinfo 2.1 only, you can however have 2.0 and 2.1 side by side, only the json for 2.1 will returns the repository key.

But that will introduce another thing to do, plume use `https://xxx/nodeinfo` so it would need to became `https://xxx/nodeinfo/version` or something like that.

keeping 2.0 would be good anyway for compatibility.

and for the schema version, if you add repository in a 2.0 version it will break on any parser forcing a validation, since 2.0 schema doesn't allow repository.

zcdunn:
Ok. I'll update that. Is it ok to make it two separate endpoints, `/nodeinfo/2.0` and `/nodeinfo/2.1` since those are the only ones we provide or should the version be parameterized?

BaptisteGelez:
Thank you! For the versions, adding a parameter in the route could indeed do the job. And I'm not sure the `json!` macro allows it, but if possible only add the repository field if the requested version is 2.1

PR #448, opened 2019-02-24
Password reset

### LINKED ISSUE #298: Password reset

BaptisteGelez:
There is now a link on the login page if you forgot your password. It redirects you to a page asking your email address. An email containing a password reset link will be sent to this address if someone registered with this address. The link is only valid for 15 minutes.

I used the `lettre` crate for mail finally since I didn't find a way to set a subject with `mailstrom` (which is very weird). It also allowed me to implement a debug mail sender, that can be enabled with `--feature debug-mailer` and that will log mails instead of actually sending them.

~~I should also document the new configuration options for the mail server.~~ Done (but I can still modify it again if you don't like the current approach)

Fixes #298

### DIFF
--- a/src/routes/session.rs
+++ b/src/routes/session.rs
@@ -1,4 +1,4 @@
-use lettre::EmailTransport;
+use lettre::Transport;
 use rocket::{
     State,
     http::{Cookie, Cookies, SameSite, uri::Uri},
@@ -162,7 +162,7 @@ pub fn password_reset_request(
             i18n!(intl.catalog, "Here is the link to reset your password: {0}"; link)
         ) {
             match *mail.lock().unwrap() {
-                Some(ref mut mail) => { mail.send(&message).map_err(|_| eprintln!("Couldn't send password reset mail")).ok(); }
+                Some(ref mut mail) => { mail.send(message.into()).map_err(|_| eprintln!("Couldn't send password reset mail")).ok(); }
                 None => {}
             }
         }
--- a/src/routes/session.rs
+++ b/src/routes/session.rs
@@ -156,14 +156,15 @@ pub fn password_reset_request(
         });
 
         let link = format!("https://{}/password-reset/{}", *BASE_URL, id);
-        let message = build_mail(
+        if let Some(message) = build_mail(
             form.email.clone(),
             i18n!(intl.catalog, "Password reset"),
             i18n!(intl.catalog, "Here is the link to reset your password: {0}"; link)
-        );
-        match *mail.lock().unwrap() {
-            Some(ref mut mail) => { mail.send(&message).map_err(|_| eprintln!("Couldn't send password reset mail")).ok(); }
-            None => {}
+        ) {
+            match *mail.lock().unwrap() {
+                Some(ref mut mail) => { mail.send(&message).map_err(|_| eprintln!("Couldn't send password reset mail")).ok(); }
+                None => {}
+            }
         }
     }
     render!(session::password_reset_request_ok(
@@ -213,10 +214,10 @@ pub fn password_reset(
     form.validate()
         .and_then(|_| {
             let mut requests = requests.lock().unwrap();
-            let req = requests.iter().find(|x| x.id == token.clone()).expect("Couldn't find the password reset request").clone();
+            let req = requests.iter().find(|x| x.id == token.clone()).ok_or(to_validation(0))?.clone();
             if req.creation_date.elapsed().as_secs() < 60 * 60 * 2 { // Reset link is only valid for 2 hours
                 requests.retain(|r| *r != req);
-                let user = User::find_by_email(&*conn, &req.mail).expect("User not found");
+                let user = User::find_by_email(&*conn, &req.mail).map_err(to_validation)?;
                 user.reset_password(&*conn, &form.password).ok();
                 Ok(Redirect::to(uri!(new: m = i18n!(intl.catalog, "Your password was successfully reset."))))
             } else {
@@ -230,4 +231,14 @@ pub fn password_reset(
                 err
             ))
         })
+}
+
+fn to_validation<T>(_: T) -> ValidationErrors {
+    let mut errors = ValidationErrors::new();
+    errors.add("", ValidationError {
+        code: Cow::from("server_error"),
+        message: Some(Cow::from("An unknown error occured")),
+        params: std::collections::HashMap::new()
+    });
+    errors
 }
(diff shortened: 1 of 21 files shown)


src/routes/session.rs
```diff
@@ -101,3 +106,112 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+pub fn password_reset(
+    conn: DbConn,
+    intl: I18n,
+    token: String,
+    requests: State<Arc<Mutex<Vec<ResetRequest>>>>,
+    form: Form<NewPasswordForm>
+) -> Result<Redirect, ErrorPage> {
+    let requests = requests.lock().unwrap();
+    let req = requests.iter().find(|x| x.id == token.clone()).ok_or(Error::NotFound)?;
+    if req.creation_date.elapsed().as_secs() < 60 * 15 { // Reset link is only valid for 15 minutes
```
fdb-hiroshima: 15 minutes seems shorts, I'd advise something like 1 or 2 hours instead, mails can be long to transmit

src/routes/session.rs
```diff
@@ -101,3 +106,112 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+    intl: I18n,
+    mail: State<Arc<Mutex<Mailer>>>,
+    form: Form<ResetForm>,
+    requests: State<Arc<Mutex<Vec<ResetRequest>>>>
+) -> Ructe {
+    if User::find_by_email(&*conn, &form.email).is_ok() {
+        let id = plume_common::utils::random_hex();
+        {
+            let mut requests = requests.lock().unwrap();
+            requests.push(ResetRequest {
```
fdb-hiroshima: to prevent the `State<...>` from growing up in memory (basically be a memory leak), you should filter and eliminate old entries either with the internal scheduler, or somewhere like here

fdb-hiroshima: you should check if there haven't already be a link sent recently, otherwise one might use it to spam a user by sending them many reset links

src/routes/session.rs
```diff
@@ -101,3 +105,139 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+
+#[post("/password-reset", data = "<form>")]
+pub fn password_reset_request(
+    conn: DbConn,
+    intl: I18n,
+    mail: State<Arc<Mutex<Mailer>>>,
+    form: Form<ResetForm>,
+    requests: State<Arc<Mutex<Vec<ResetRequest>>>>
+) -> Ructe {
+    if User::find_by_email(&*conn, &form.email).is_ok() && requests.lock().unwrap().iter().find(|x| x.mail == form.email.clone()).is_none() {
```
fdb-hiroshima: You should maintain the lock instead of locking and unlocking many times

fdb-hiroshima: I think using ! .any(|x|...) would be easier to read

src/routes/session.rs
```diff
@@ -101,3 +105,139 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+        {
+            let mut requests = requests.lock().unwrap();
+            requests.push(ResetRequest {
+                mail: form.email.clone(),
+                id: id.clone(),
+                creation_date: Instant::now(),
+            });
+        }
+        // Remove outdated requests (more than 1 day old) to avoid the list to grow too much
+        let to_remove = {
```
fdb-hiroshima: you should remove before checking if there is already an entry, otherwise one asking for a second reset link after, says, 12 hours, won't get it's mail. Also this mean the vec is cleaned only when a valid request is made, so if nobody else request a password reset, one wont ever get unlocked

fdb-hiroshima: Instead of building a Vec, then "substracting" if from the one stored, building directly the right one and mem::replace-ing it seems easier to read, and don't require to enable another feature
(snippet : 
```rust
let a = Mutex::new(vec![1,2,3]);
std::mem::replace(&mut *a.lock().unwrap(), vec![4,5,6]);
```

src/routes/session.rs
```diff
@@ -101,3 +105,139 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+            let lock = requests.lock().unwrap();
+            lock.clone().into_iter()
+                .filter(|r| r.creation_date.elapsed().as_secs() > 24 * 60)
```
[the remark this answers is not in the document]
BaptisteGelez: Oops looks like I don't know how time is working :grimacing:

src/routes/session.rs
BaptisteGelez: I can't because I need to borrow it both mutably and immutably and Rust won't let me do that. But maybe there is a solution to do it I didn't find?

fdb-hiroshima: I've removed a lot of code, but here is a poc of how I would do it (and some other things, trashing old things sooner and don't use the feature) https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=2c32af184c99ff62ee99f94aa2e4fa13
Sadly I can't compile this branch so I can't give you a clean commit, but it should put you on the right track

src/routes/session.rs
fdb-hiroshima: (I found a better way to do it, see the link to the playground)

src/routes/session.rs
```diff
@@ -101,3 +105,129 @@ pub fn delete(mut cookies: Cookies) -> Redirect {
...
+pub fn password_reset_request(
+    conn: DbConn,
+    intl: I18n,
+    mail: State<Arc<Mutex<Mailer>>>,
+    form: Form<ResetForm>,
+    requests: State<Arc<Mutex<Vec<ResetRequest>>>>
+) -> Ructe {
+    let mut requests = requests.lock().unwrap();
+    // Remove outdated requests (more than 1 day old) to avoid the list to grow too much
+    requests.retain(|r| r.creation_date.elapsed().as_secs() < 24 * 60 * 6);
```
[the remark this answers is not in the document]
igalic: why is math so hard? why are there no `as_minutes()` and `as_hours()` functions?!??

fdb-hiroshima:
It doesn't seems to compile with debug-mailer on

BaptisteGelez:
Already done :grin:, see https://docs.joinplu.me/environment/ and https://docs.joinplu.me/installation/config/

fdb-hiroshima:
Mb, I did not check it before. I think however Docker documentation should be updated, as it says only 2 vars need changes, whereas MAIL_* do too

PR #452, opened 2019-02-26
Slightly improve the media experience

### LINKED ISSUE #432: Image description links in https://<url>/medias are not shown properly.
When logged in your can view your uploaded images in https://<url>/medias. When the screen resolution is high enough / browser window is wide enough, the page displays okay. But I'm using a 1440x900 screen and it's not wide enough to display the page quite right.

The images are displayed too large so that the image descriptions to their right are not always shown. This means you can't access the images' info pages where the text to copy to add the posts is shown.

The design of the https://<url>/medias page should be improved so the image descriptions are always shown, including on mobile. The display is very poor on mobile right now.

Thanks!

BaptisteGelez:
- Use a grid to display the list of media
- Add icons for non-image media preview
- Paginate the gallery
- Add links to the gallery in the editor and in the profile settings to make it more discoverable when you need it

Here are some screenshots:

screenshot_2019-02-26 vos medias plume
screenshot_2019-02-26 nouvel article plume
screenshot_2019-02-26 modifier votre compte plume

Fixes #432


plume-models/src/safe_string.rs
```diff
@@ -19,13 +19,23 @@ lazy_static! {
...
+            .add_tag_attributes(
+                "video",
+                [ "src", "title", "controls" ].iter(),
+            )
+            .add_tag_attributes(
+                "audio",
+                [ "src", "title", "controls" ].iter(),
+            )
+            .add_tag_attributes("div", [ "class" ].iter())
+            .add_tag_attributes("a", [ "class" ].iter());
```
fdb-hiroshima: I'm not a big fan off allowing custom class, peoples could inject div/a with the right classes to mess up the interface. I'd prefer maybe an "unsafe" (not necessary in the sens of rust's unsafe) builder for SafeString, with the contract that things put in it must already be trusted

plume-models/src/medias.rs
```diff
@@ -102,19 +102,19 @@ impl Media {
     pub fn html(&self, conn: &Connection) -> Result<SafeString> {
         let url = self.url(conn)?;
         Ok(match self.category() {
-            MediaCategory::Image => SafeString::new(&format!(
+            MediaCategory::Image => SafeString::trusted(&format!(
```
fdb-hiroshima: the alt and title can't be trusted (and trusted is unnecessary here)

plume-models/src/safe_string.rs
```diff
@@ -25,17 +25,7 @@ lazy_static! {
...
                 "iframe",
                 [ "width", "height", "src", "frameborder" ].iter().map(|&v| v),
-            )
-            .add_tag_attributes(
-                "video",
-                [ "src", "title", "controls" ].iter(),
-            )
-            .add_tag_attributes(
-                "audio",
-                [ "src", "title", "controls" ].iter(),
```
fdb-hiroshima: I think we should keep video and audio tags, So that people can insert media content in their posts

plume-models/src/medias.rs
```diff
@@ -102,19 +102,19 @@ impl Media {
...
                 r#"<img src="{}" alt="{}" title="{}">"#,
                 url, escape(&self.alt_text), escape(&self.alt_text)
             )),
-            MediaCategory::Audio => SafeString::new(&format!(
+            MediaCategory::Audio => SafeString::trusted(&format!(
```
fdb-hiroshima: title can't be trusted. I don't know how to handle this one. Maybe have some way of concatenating SafeString, and do `concat(trusted("<div></div>"),new("<audio/>"))`

plume-models/src/medias.rs
```diff
@@ -102,19 +102,19 @@ impl Media {
...
                 r#"<div class="media-preview audio"></div><audio src="{}" title="{}" controls></audio>"#,
                 url, escape(&self.alt_text)
             )),
-            MediaCategory::Video => SafeString::new(&format!(
+            MediaCategory::Video => SafeString::trusted(&format!(
```
fdb-hiroshima: title can't be trusted (and would be unnecessary here if we keep video/audio allowed)

PR #455, opened 2019-02-27
Fix follow IDs

### LINKED ISSUE #449: Strange urls in follows.ap_url column
URLs insterted in the `follows.ap_url` column have a strange structure:
```
plume=# select ap_url from follows;
                                                       ap_url                                                       
--------------------------------------------------------------------------------------------------------------------
 
 https://101010.pl/users/EuropaMiasto/follow/https://blog.funkwhale.audio/@/funkwhale/
 https://3.distsn.org/users/vaginaplant/follow/https://baptiste.gelez.xyz/@/vaginaplant/
 https://3.distsn.org/users/vaginaplant/follow/https://blog.funkwhale.audio/@/funkwhale/
 https://3.distsn.org/users/vaginaplant/follow/https://fediverse.blog/@/rinsuki/
 https://abyss.fun/users/guskma/follow/https://plume.mastodon.host/@/guskma/
 https://aleph.land/users/yahananxie/follow/https://fediverse.blog/@/yuki/
 https://anticapitalist.party/users/a000d4f7a91939d0e71df1646d7a48/follow/https://fediverse.blog/@/fdb-hiroshima/
 https://anticapitalist.party/users/a000d4f7a91939d0e71df1646d7a48/follow/https://fediverse.blog/@/FDB_hiroshima/
 https://anticapitalist.party/users/a000d4f7a91939d0e71df1646d7a48/follow/https://fediverse.blog/@/Luna/
 https://anticapitalist.party/users/Cocoron/follow/https://cafe.sunbeam.city/@/Cocoron/
```

As you can see, it looks like the URL is made of two concatenated URLs.

Plume version (and/or commit): latest docker tag

BaptisteGelez:
I changed the syntax of the `insert!` macro, allowing to modify an object just after it was saved to the database. You can now optional pass it a "closure" (actually it just looks like a closure) that will be run just after the insertion.

I updated all the models, including Follow, to use it to generate AP ID and URL. The advantages are that the ID/URL are now correct (fixes #449) and that we don't have methods like `update_boxes` or `update_ap_url` to call manually and that were easy to forget.

### DIFF
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -1,6 +1,6 @@
 use activitypub::{actor::Group, collection::OrderedCollection, Actor, CustomObject, Object};
 use chrono::NaiveDateTime;
-use diesel::{self, ExpressionMethods, QueryDsl, RunQueryDsl};
+use diesel::{self, ExpressionMethods, QueryDsl, RunQueryDsl, SaveChangesDsl};
 use openssl::{
     hash::MessageDigest,
     pkey::{PKey, Private},
@@ -30,7 +30,7 @@ use {Connection, BASE_URL, USE_HTTPS, Error, Result};
 
 pub type CustomGroup = CustomObject<ApSignature, Group>;
 
-#[derive(Queryable, Identifiable, Serialize, Deserialize, Clone)]
+#[derive(Queryable, Identifiable, Serialize, Deserialize, Clone, AsChangeset)]
 pub struct Blog {
     pub id: i32,
     pub actor_id: String,
@@ -45,7 +45,7 @@ pub struct Blog {
     pub public_key: String,
 }
 
-#[derive(Insertable)]
+#[derive(Default, Insertable)]
 #[table_name = "blogs"]
 pub struct NewBlog {
     pub actor_id: String,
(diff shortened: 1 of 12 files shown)


src/routes/posts.rs
```diff
@@ -227,7 +227,6 @@ pub fn update(blog: String, slug: String, user: User, cl: ContentLen, form: Leni
             post.license = form.license.clone();
             post.cover_id = form.cover;
             post.update(&*conn, &searcher).expect("post::update: update error");;
-            let post = post.update_ap_url(&*conn).expect("post::update: update ap url error");
```
BaptisteGelez: Just a note: I removed that without replacing it with something else, since it is incorrect (the AP IDs should change), and was not working anyway (since we were checking if `ap_url` was empty before updating it, which was only the case on newly created articles).

plume-models/src/blogs.rs
```diff
@@ -431,12 +427,10 @@ impl NewBlog {
             actor_id,
             title,
             summary,
-            outbox_url: String::from(""),
-            inbox_url: String::from(""),
             instance_id,
-            ap_url: String::from(""),
             public_key: String::from_utf8(pub_key).or(Err(Error::Signature))?,
             private_key: Some(String::from_utf8(priv_key).or(Err(Error::Signature))?),
+            ..NewBlog::default()
```
fdb-hiroshima: I've never used this syntax, but it looks super nice

plume-models/src/follows.rs
```diff
@@ -35,7 +35,14 @@ pub struct NewFollow {
 }
 
 impl Follow {
-    insert!(follows, NewFollow);
+    insert!(follows, NewFollow, |inserted, conn| {
+        if inserted.ap_url.is_empty() {
+            inserted.ap_url = ap_url(&format!("{}/follows/{}", *BASE_URL, inserted.id));
```
fdb-hiroshima: this route should really get implemented one day

BaptisteGelez: Yes, I think we already have an issue for it…

plume-models/src/blogs.rs
igalic: what does it do?

BaptisteGelez: For all the fields that were not provided, it takes their value from the given struct (here `NewBlog::default()`).

PR #457, opened 2019-03-05
Add a fqn field to blogs and users

### LINKED ISSUE #319: Cache FQNs in DB
We are recomputing them so many time when they would just be stored as regular field in the database.

BaptisteGelez:
Fixes #319

### DIFF
--- a/migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
+++ b/migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
@@ -1,0 +1,3 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE blogs DROP COLUMN fqn;
+ALTER TABLE users DROP COLUMN fqn;
--- a/migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
+++ b/migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
@@ -1,3 +1,46 @@
 -- This file should undo anything in `up.sql`
-ALTER TABLE blogs DROP COLUMN fqn;
-ALTER TABLE users DROP COLUMN fqn;
+CREATE TABLE blogs_no_fqn (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    actor_id VARCHAR NOT NULL,
+    title VARCHAR NOT NULL,
+    summary TEXT NOT NULL DEFAULT '',
+    outbox_url VARCHAR NOT NULL UNIQUE,
+    inbox_url VARCHAR NOT NULL UNIQUE,
+    instance_id INTEGER REFERENCES instances(id) ON DELETE CASCADE NOT NULL,
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url text not null default '' UNIQUE,
+    private_key TEXT,
+    public_key TEXT NOT NULL DEFAULT '', fqn TEXT NOT NULL DEFAULT '',
+    CONSTRAINT blog_unique UNIQUE (actor_id, instance_id)
+);
+
+INSERT INTO blogs_no_fqn SELECT * FROM blogs;
+DROP TABLE blogs;
+ALTER TABLE blogs_no_fqn RENAME TO blogs;
+
+
+CREATE TABLE users_no_fqn (
+    id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+    username VARCHAR NOT NULL,
+    display_name VARCHAR NOT NULL DEFAULT '',
+    outbox_url VARCHAR NOT NULL UNIQUE,
+    inbox_url VARCHAR NOT NULL UNIQUE,
+    is_admin BOOLEAN NOT NULL DEFAULT 'f',
+    summary TEXT NOT NULL DEFAULT '',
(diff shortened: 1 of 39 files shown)


fdb-hiroshima:
I'm wondering. On a brand new, uninitialized instance, does it works? The select would return 0 rows, so does the CASE never match, or does it fail?

BaptisteGelez:
I think it will never run, since there are no blogs or users, so no problem.

migrations/sqlite/2019-03-05-082846_add_fqn/up.sql
```diff
@@ -1,7 +1,7 @@
 -- Your SQL goes here
 ALTER TABLE blogs ADD COLUMN fqn TEXT NOT NULL DEFAULT '';
 UPDATE blogs SET fqn =
-	(CASE WHEN instance_id = 1 THEN
+	(CASE WHEN (SELECT local FROM instances WHERE id = instance_id) THEN
 		actor_id
 	ELSE
 		(actor_id || '@' || (SELECT public_domain FROM instances WHERE id = instance_id LIMIT 1))
```
fdb-hiroshima: why LIMIT 1? Isn't id unique?

BaptisteGelez: Just to be sure. :stuck_out_tongue:

migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
```diff
@@ -0,0 +1,3 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE blogs DROP COLUMN fqn;
+ALTER TABLE users DROP COLUMN fqn;
```
fdb-hiroshima: sqlite doesn't allow to drop column, you have to do the annoying copy thing

BaptisteGelez: Ah, yes, forgot about that… :expressionless:

migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
```diff
@@ -1,3 +1,46 @@
...
+    actor_id VARCHAR NOT NULL,
+    title VARCHAR NOT NULL,
+    summary TEXT NOT NULL DEFAULT '',
+    outbox_url VARCHAR NOT NULL UNIQUE,
+    inbox_url VARCHAR NOT NULL UNIQUE,
+    instance_id INTEGER REFERENCES instances(id) ON DELETE CASCADE NOT NULL,
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url text not null default '' UNIQUE,
+    private_key TEXT,
+    public_key TEXT NOT NULL DEFAULT '', fqn TEXT NOT NULL DEFAULT '',
```
fdb-hiroshima: `blogs_no_fqn`, `fqn TEXT NOT NULL`...

migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
```diff
@@ -1,3 +1,46 @@
...
+    hashed_password TEXT,
+    instance_id INTEGER REFERENCES instances(id) ON DELETE CASCADE NOT NULL,
+    creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    ap_url TEXT NOT NULL default '' UNIQUE,
+    private_key TEXT,
+    public_key TEXT NOT NULL DEFAULT '',
+    shared_inbox_url VARCHAR,
+    followers_endpoint VARCHAR NOT NULL DEFAULT '' UNIQUE,
+    avatar_id INTEGER REFERENCES medias(id) ON DELETE SET NULL,
+    last_fetched_date TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, fqn TEXT NOT NULL DEFAULT '',
```
fdb-hiroshima: `users_no_fqn`, `fqn TEXT NOT NULL`...

migrations/sqlite/2019-03-05-082846_add_fqn/down.sql
BaptisteGelez: Ah, that's why it was telling me I was not supplying enough values when doing `INSERT INTO blogs_no_fqn SELECT all, the, columns, but, not, fqn FROM blogs;`. :grimacing:

plume-models/src/schema.rs
```diff
@@ -47,160 +47,160 @@ table! {
     }
 }
 
+table! {
+    comment_seers (id) {
+        id -> Nullable<Integer>,
```
fdb-hiroshima: this is not supposed to be Nullable, but I don't get where it come from, your migrations aren't touching this table

BaptisteGelez: Diesel is regenerating the schema when running migrations, so it changed when I tested with SQlite.

fdb-hiroshima: For some reason, comment_seers's id is an `INTEGER PRIMARY KEY AUTOINCREMENT`, not an `INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT` on sqlite (but fine on pg I assume), that's why the scheme was invalid

PR #458, opened 2019-03-05
New editor

### LINKED ISSUE #255: New visual editor
A new WYSIWYG editor should be added for people who don't know markdown. It should allow you to focus on writing. It will probably need to add some API endpoints for synchronization with the server.

A first version can be found here: https://github.com/Plume-org/Plume/tree/new-editor

And according to this [vote](https://framavox.org/p/SjZHuPWh/what-to-do-with-the-current-editor-), the current editor should be kept.

BaptisteGelez:
With this PR, when JS is activated and WASM supported, the article editor will be dynamically replaced with `contenteditable`s elements. This makes the editing interface simpler and less like a regular form. It will also allow us to easily add visual formatting with native browser APIs (and to insert images or videos directly). Here is a little demo:

peek 05-03-2019 16-12

There is still a lot to do, but it is a good first step.

Fixes #255


BaptisteGelez:
Something I just noticed: published articles are always drafts (unless you are editing an already published article)

fdb-hiroshima:
I'm having trouble making it word. I had to tinker plume-front.js. Any idea why, or how to prevent that?

BaptisteGelez:
Was the build script correctly run by cargo? Like, did you saw `plume(build)` before `plume(bin)` next to the compilation progress bar? Otherwise, you can try to add a `println!("cargo:rerun-if-changed=target/deploy/plume-front.js")` (IIRC) to it, so that we are sure it is re-run when the JS changes (it does when the WASM changes normally, but maybe that's not enough).

fdb-hiroshima:
ok my bad, I ran `cargo web deploy -o ../static` so the buildscript did not update the js file as it should have

plume-front/src/editor.rs
```diff
@@ -0,0 +1,208 @@
...
+                        let button = document().create_element("input").unwrap();
+                        js!{
+                            @{&button}.type = "submit";
+                            @{&button}.value = @{i18n!(CATALOG, "Publish")};
+                        };
+                        button.append_child(&document().create_text_node(&i18n!(CATALOG, "Publish")));
+                        button.add_event_listener(mv!(widgets, old_ed => move |_: ClickEvent| {
+                            set_value("title", widgets.0.inner_text());
+                            set_value("subtitle", widgets.1.inner_text());
+                            set_value("editor-content", widgets.2.inner_text());
```
fdb-hiroshima: By doing the following change, I was able to copy past a picture in a post and have it rendered as expected. However when re-editing the article, `<` and `>` where escaped, leading to a messy html tag in the middle of the content. I don't know if this is considered in scope for this pr, but this might be something to look at, as it will get important once the editor get richer

```suggestion
                            set_value("editor-content", js!{return @{&widgets.2}.innerHTML}.as_str().unwrap_or_default());
```

po/plume-front/en.po~
```diff
@@ -0,0 +1,29 @@
+msgid ""
```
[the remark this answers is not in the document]
igalic: my suggestion to @BaptisteGelez is to either add this our `.gitignore` or to her global `.git/ignore`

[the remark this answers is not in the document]
BaptisteGelez: No, I though `*.po~` was in the gitignore, but it looks like it is not. :woman_shrugging: I will add it.

plume-front/src/editor.rs
BaptisteGelez: I think it would be easy to avoid escaping here, I will try.

po/plume-front/en.po~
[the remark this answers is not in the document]
fdb-hiroshima: They are only ignored in some directories, not everywhere

BaptisteGelez:
I added back the character counter, but it is not as precise as the previous one, since I only took the article content in account (otherwise I would have had to rebuild all the original form every time, which would be quite long both in code and when running). I don't think that's a big issue, since the previous one was not perfectly precise neither, and that people will rarely reach this limit anyway.

.gitignore
```diff
@@ -3,8 +3,8 @@ rls
 **/*.rs.bk
 rls
 translations
-po/*.po~
-po/plume/*.po~
+*.po~
+*.po~
```
fdb-hiroshima: this line is duplicated

plume-front/src/editor.rs
```diff
@@ -0,0 +1,244 @@
...
+
+fn no_return(evt: KeyDownEvent) {
+    if evt.key() == "Enter" {
+        evt.prevent_default();
+    }
+}
+
+pub fn init() {
+    document().get_element_by_id("plume-editor")
+        .and_then(|ed| {
```
fdb-hiroshima: this closure is really long. It would gain in readability to be divided in multiple smaller functions called one after an other (like "hide non-js form", "add js editor", "show popup", "copy editor to form"...). I understand what it's doing, but if there is something wrong, it's too complex to see it right away

plume-front/src/main.rs
```diff
@@ -1,56 +1,35 @@
...
+
+init_i18n!("plume-front", en, fr);
+
+mod editor;
+
+compile_i18n!();
+
+lazy_static! {
+    static ref CATALOG: gettext::Catalog = {
+        let catalogs = include_i18n!();
```
fdb-hiroshima: I love that it just works like in the back-end, with nothing more to

BaptisteGelez: Yeah, that's really great! The only downside is that we include all the translations in the .wasm so it is a bit bigger, but there is like 3 or 4 messages for the moment so I guess it's fine (and I don't expect it to become much bigger, I think we will never have more than 30 messages for the front end).

fdb-hiroshima: considering all Plume translations (not plume-front) are 208k on my computer, any picture loaded on the front page must cost much more than having all translations available

BaptisteGelez:
I tried to make the code a bit more readable, but I'm not sure it is enough…

plume-front/src/main.rs
igalic: so how big is the dynamic front-end right now? and how much of that is cachable?

BaptisteGelez: It is 2.2M on my computer, but there are various ways to optimize WASM size that we didn't explored yet. And I don't know how much of it is cachable? Nothing?

fdb-hiroshima: I'd say everything, it's a static file after all. I tried compiling with `cargo web deploy --release`, it seems to work, and is "only" 832k. We could also try to compact plume-front.js as it's 35k with indentation and things, however I'm not sure it's that important

fdb-hiroshima:
Oups, 2 little things : It doesn't seems to be possible to add return carriage in the content (prevent default I guess), and it may be nice to have a button to switch back to the plain markdown editor

PR #459, opened 2019-03-06
Make media extension parsing safer

BaptisteGelez:
Only keep it if contains letters and numbers only, otherwise remove it.

To be merged before #452


igalic:
How does Rocket handle the mime type of a file that has no extension?
or is determining the MIME type (always) left as an exercise to the proxy server?

BaptisteGelez:
It looks like the header is not set when no extension is found (or when Rocket doesn't know what the associated MIME type is) https://github.com/SergioBenitez/Rocket/blob/master/core/lib/src/response/named_file.rs#L84

PR #460, opened 2019-03-06
Edit blogs, and add blog icons and banners

### LINKED ISSUE #453: blog summary / blog update missing
Currently both, the database schema, and the templates, mention a blog summary

however, when creating a new blog, there's no such field

furthermore, there's no option to add it later, since there's no `update` for blogs

### LINKED ISSUE #454: Blog icon/avatar and banner
We should be able to select a blog avatar/icon and a banner for a blog. However this would probably require adding another file upload endpoint in the blog settings, and I don't think it is a good idea until there is support for file upload directly in Rocket, as the code we currently have to handle that is quite big and may be hard to make more generic.

BaptisteGelez:
Also adds a parameter to `md_to_html` to only render inline elements (so that we don't have titles or images in blog descriptions). And moves the delete button for the blog on the edition page.

I still have to update the SQLite migration once others PRs with migrations will be merged.

Also, there will be a problem when you edit a blog while not owning its banner or icon: when validating they will be reset to their default values… I don't see a good solution to this until we have a better way to handle uploads with Rocket (the same is probably happening for articles btw).

And the icon/banner are not federated yet, I don't know if I should add it to this PR or if it can come after?

image
image

Fixes #453
Fixes #454


plume-models/src/blogs.rs
```diff
@@ -236,6 +243,10 @@ impl Blog {
                     .public_key_publickey()?
                     .public_key_pem_string()?,
                 private_key: None,
+                // TODO: federate the two following properties
```
igalic: or all three?

templates/blogs/details.rs.html
```diff
@@ -19,17 +19,36 @@
...
-    </p>
+    @if blog.banner_id.is_some() {
+        <div class="cover" style="background-image: url('@Html(blog.banner_url(ctx.0).unwrap_or_default())')"></div>
+        <img class="hidden u-photo" src="@blog.banner_url(ctx.0).unwrap_or_default()"/>
+    }
+    <div class="h-card">
+        <div class="user">
+            <div class="flex wrap">
+                <div class="avatar medium" style="background-image: url('@blog.icon_url(ctx.0)');" aria-label="@i18n!(ctx.1, "{}'s icon"; &blog.title)"></div>
+                <img class="hidden u-photo" src="@blog.icon_url(ctx.0)"/>
```
igalic: isn't the Avatar also optional?

fdb-hiroshima: icon_url returns either the blog avatar, or the default one (see https://github.com/Plume-org/Plume/pull/460/files#diff-a4a32b57517e9459ff561610609f8044R364 )

plume-models/src/blogs.rs
BaptisteGelez: Yes, at the time I wrote the comment `summary_html` was not here yet. :grin:

templates/blogs/details.rs.html
igalic: duh, thanks!

src/routes/blogs.rs
```diff
@@ -136,6 +137,76 @@ pub fn delete(conn: DbConn, name: String, user: Option<User>, intl: I18n, search
...
+    let mut blog = Blog::find_by_fqn(&*conn, &name).expect("blog::update: blog not found");
+    if user.clone().and_then(|u| u.is_author_in(&*conn, &blog).ok()).unwrap_or(false) {
+        let user = user.expect("blogs::edit: User was None while it shouldn't");
+        form.validate()
+            .and_then(|_| {
+                blog.title = form.title.clone();
+                blog.summary = form.summary.clone();
+                blog.summary_html = SafeString::new(&utils::md_to_html(&form.summary, "", true).0);
+                blog.icon_id = form.icon;
+                blog.banner_id = form.banner;
```
fdb-hiroshima: user can provide the id of a media they don't own, or of a non-picture media

templates/blogs/details.rs.html
```diff
@@ -19,17 +19,37 @@
...
-    <p>
-        @i18n!(ctx.1, "There's one author on this blog: ", "There are {0} authors on this blog: "; authors.len())
-        @for author in authors {
-            <a class="author p-author" href="@uri!(user::details: name = &author.fqn)">@author.name()</a>
-        }
-    </p>
-    <p>
-        @i18n!(ctx.1, "There's one article on this blog", "There are {0} articles on this blog"; total_articles)
-    </p>
+    @if blog.banner_id.is_some() {
```
fdb-hiroshima: `if let Some(banner) = blog.banner_url(ctx0)` would make more sense here, so there is no unwrap_or_default

plume-models/src/blogs.rs
```diff
@@ -346,6 +432,16 @@ impl Blog {
         })
     }
 
+    pub fn icon_url(&self, conn: &Connection) -> String {
+        self.icon_id.and_then(|id|
+            Media::get(conn, id).and_then(|m| m.url(conn)).ok()
+        ).unwrap_or("/static/default-avatar.png".to_string())
+    }
+
+    pub fn banner_url(&self, conn: &Connection) -> Option<String> {
```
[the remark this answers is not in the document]
igalic: not if my pr gets merged first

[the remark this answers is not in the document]
BaptisteGelez: It makes sense to me that they don't have the same return type: in one case we provide a fallback URL, in the other case we just want to not display any image.

And I don't know what would be the correct way to replace this `unwrap_or`… :sweat_smile:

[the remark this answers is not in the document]
fdb-hiroshima: `unwrap_or_else` because you're calling a function, so let's not call it if it's not necessary

src/routes/blogs.rs
```diff
@@ -149,6 +151,102 @@ pub fn delete(name: String, rockets: PlumeRocket) -> Result<Redirect, Ructe>{
...
+        // TODO actually return 403 error code
+        Ok(render!(errors::not_authorized(
+            &(&*conn, &intl.catalog, user),
+            i18n!(intl.catalog, "You are not allowed to edit this blog.")
+        )))
+    }
+}
+
+/// Returns true if the media is owned by `user` and is a picture
+fn check_media(conn: DbConn, id: i32, user: &User) -> bool {
```
fdb-hiroshima: Does this function actually do what it says it does?

BaptisteGelez: Looks like I pushed too fast… :sweat_smile:

fdb-hiroshima:
There is no way to remove pictures once you set theme, you can just replace with something else

plume-models/src/blogs.rs
```diff
@@ -32,6 +32,7 @@ use {Connection, Error, Result, CONFIG};
 pub type CustomGroup = CustomObject<ApSignature, Group>;
 
 #[derive(Queryable, Identifiable, Clone, AsChangeset)]
+#[changeset_options(treat_none_as_null = "true")]
```
fdb-hiroshima: I don't understand what this does, but apparently it did the trick

BaptisteGelez: It is an option for the `#[derive(AsChangeset)]`, that tells diesel to consider `Option` fields that are `None` to be considered as `NULL` in SQL, instead of not being included in the query at all.

fdb-hiroshima: Ok, doesn't this potentially break other things that relied on it before in our code?

igalic: not according to our extensive tests

fdb-hiroshima: Test in production :tm:

BaptisteGelez: > Ok, doesn't this potentially break other things that relied on it before in our code?

It only affects the `AsChangeset` implementation, which is used when `UPDATE`ing blogs, so I think it is the only place we are doing that.

PR #462, opened 2019-03-08
Refactor with the help of Clippy

igalic:
We add clippy as our build — also rectifying the missing `plume-cli` build!

In the next step we follow clippy's advise and fix some of the "simple" mistakes in our code, such as style or map usage.

Finally, we refactor some hard bits that need extraction of new types, or refactoring of function call-types, especially those that thread thru macros, and, of course functions with ~15 parameters should probably be rethought.

### DIFF
--- a/plume-models/src/search/query.rs
+++ b/plume-models/src/search/query.rs
@@ -208,7 +208,7 @@ impl PlumeQuery {
     }
 
     // split a string into a token and a rest
-    pub fn get_first_token<'a>(mut query: &'a str) -> (&'a str, &'a str) {
+    pub fn get_first_token(mut query: &str) -> (&str, &str) {
         query = query.trim();
         if query.is_empty() {
             ("", "")
@@ -244,22 +244,22 @@ impl PlumeQuery {
     fn from_str_req(&mut self, mut query: &str) -> &mut Self {
         query = query.trim_left();
         if query.is_empty() {
-            self
-        } else {
-            let occur = if query.get(0..1).map(|v| v=="+").unwrap_or(false) {
-                query = &query[1..];
-                Occur::Must
-            } else if query.get(0..1).map(|v| v=="-").unwrap_or(false) {
-                query = &query[1..];
-                Occur::MustNot
-            } else {
-                Occur::Should
-            };
-            gen_parser!(self, query, occur; normal: title, subtitle, content, tag,
-                            instance, author, blog, lang, license;
-                            date: after, before);
-            self.from_str_req(query)
+            return self
         }
+
+        let occur = if query.get(0..1).map(|v| v=="+").unwrap_or(false) {
+            query = &query[1..];
+            Occur::Must
+        } else if query.get(0..1).map(|v| v=="-").unwrap_or(false) {
+            query = &query[1..];
+            Occur::MustNot
+        } else {
+            Occur::Should
+        };
+        gen_parser!(self, query, occur; normal: title, subtitle, content, tag,
+                        instance, author, blog, lang, license;
+                        date: after, before);
+        self.from_str_req(query)
     }
 
     // map a token and it's field to a query
--- a/plume-models/src/search/query.rs
+++ b/plume-models/src/search/query.rs
@@ -226,26 +226,22 @@ impl PlumeQuery {
         query = query.trim();
         if query.is_empty() {
             ("", "")
-        } else {
-            if query.get(0..1).map(|v| v=="\"").unwrap_or(false) {
-                    if let Some(index) = query[1..].find('"') {
-                        query.split_at(index+2)
-                    } else {
-                        (query, "")
-                    }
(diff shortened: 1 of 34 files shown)


fdb-hiroshima:
(Most functions with ~15 parameters are routes, we won't be able to fix these warnings)

igalic:
couldn't we extract some of those parameters into a Struct?

fdb-hiroshima:
We could create a struct containing `DbConn`, `Worker`, `I18n`, `Searcher` and `Option<User>`(current connected user). Most routes requires at least 3 of these, so requires all

igalic:
> We could create a struct containing `DbConn`, `Worker`, `I18n`, `Searcher` and `Option<User>`(current connected user). Most routes requires at least 3 of these, so requires all

I'm thinking of this:

```rust
pub struct PlumeRocket {
    slug: String,
    conn: DbConn,
    user: Option<User>,
    intl: I18n,
    worker: Option<Worker>,
    searcher: Option<Searcher>,
}
```

i'm not sure about the name of the `struct` yet (or where to put it)

src/routes/blogs.rs
```diff
@@ -17,19 +17,20 @@ use plume_models::{
...
-pub fn details(intl: I18n, name: String, conn: DbConn, user: Option<User>, page: Option<Page>) -> Result<Ructe, ErrorPage> {
+pub fn details(name: String, page: Option<Page>, rockets: PlumeRocket) -> Result<Ructe, ErrorPage> {
     let page = page.unwrap_or_default();
+    let conn = rockets.conn;
     let blog = Blog::find_by_fqn(&*conn, &name)?;
     let posts = Post::blog_page(&*conn, &blog, page.limits())?;
     let articles_count = Post::count_for_blog(&*conn, &blog)?;
     let authors = &blog.list_authors(&*conn)?;
+    let user = rockets.user;
+    let intl = rockets.intl;
```
BaptisteGelez: Maybe Rust supports destructuring (I don't remember to be honest, so you could try to do:

```rust
let { user, intl } = rockets;
```

src/routes/blogs.rs
```diff
@@ -82,8 +87,11 @@ fn valid_slug(title: &str) -> Result<(), ValidationError> {
 }
 
 #[post("/blogs/new", data = "<form>")]
-pub fn create(conn: DbConn, form: LenientForm<NewBlogForm>, user: User, intl: I18n) -> Result<Redirect, Ructe> {
+pub fn create(form: LenientForm<NewBlogForm>, rockets: PlumeRocket) -> Result<Redirect, Ructe> {
     let slug = utils::make_actor_id(&form.title);
+    let conn = rockets.conn;
+    let intl = rockets.intl;
+    let user = rockets.user.unwrap();
```
BaptisteGelez: Maybe in another PR (because I think it is a bit out of scope for this one), we could replace all the "login required" routes with something like:

```rust
rockets.user.ok_or(Err(ErrorPage::Unauthorized("You need to be logged in to do this")))?;
```

(I don't remember if `ok_or` is the correct function to use here, but I think you got what I wanted to do)

plume-models/src/search/query.rs
```diff
@@ -259,25 +241,28 @@ impl PlumeQuery {
     }
 
     // recursive parser for query string
+    // allow this clippy lint for now, until someone figures out how to
+    // refactor this better.
+    #[allow(clippy::wrong_self_convention)]
```
fdb-hiroshima: seeing `parse_query`, I think this function could be moved into it, and then we can call `parse_query` instead of `from_str_req`

src/routes/blogs.rs
fdb-hiroshima: Your syntax is not right, but there is a way https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=572b846e985d4780607766e40563e6ba

src/routes/mod.rs
```diff
@@ -5,9 +5,45 @@ use rocket::{
...
+    worker: Option<Worker<'a>>,
+}
+
+impl<'a, 'r> FromRequest<'a, 'r> for PlumeRocket<'a> {
+    type Error = ();
+
+    fn from_request(request: &'a Request<'r>) -> request::Outcome<PlumeRocket<'a>, ()> {
+        let conn = request.guard::<DbConn>()?;
+        let intl = request.guard::<I18n>()?;
+        let user = Some(request.guard::<User>()?);
```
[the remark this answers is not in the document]
fdb-hiroshima: up

src/routes/mod.rs
```diff
@@ -1,13 +1,50 @@
...
+
+use Worker;
+use Searcher;
+
+pub struct PlumeRocket<'a> {
+    conn: DbConn,
+    intl: I18n,
+    searcher: Option<Searcher<'a>>,
+    user: Option<User>,
+    worker: Option<Worker<'a>>,
```
fdb-hiroshima: These could be `Searcher<'a>` and `Worker<'a>` instead of `Option<...>`

PR #480, opened 2019-03-15
Caching

fdb-hiroshima:
Try to cache static resources.
Static non media file's timestamps are hashed at compile tome so that modifying them invalidate old cache.
~~For some reason, plume-front.wasm isn't loaded from cache, even if it contains the right headers.~~ 
I changed nothing and now it is? I guess I'm ok with it
~~I'll look into adding Etag support too~~
All response made from templates and not containing a form are now Etag-cached

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,8 +1,39 @@
-extern crate plume_common;
 extern crate ructe;
 extern crate rsass;
 use ructe::*;
 use std::{env, fs::*, io::Write, path::PathBuf};
+use std::process::{Command, Stdio};
+
+fn compute_static_hash() -> String {
+    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat --printf='%n %Y\n' | sha256sum"
+
+    let find = Command::new("find")
+        .args(&["static/", "-type", "f", "!", "-path", "static/media/*"])
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed find command");
+
+    let sort = Command::new("sort")
+        .stdin(find.stdout.unwrap())
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed sort command");
+
+    let xargs = Command::new("xargs")
+        .args(&["stat", "--printf='%n %Y\n'"])
+        .stdin(sort.stdout.unwrap())
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed xargs command");
+
+    let sha = Command::new("sha256sum")
+        .stdin(xargs.stdout.unwrap())
+        .output()
+        .expect("failed sha256sum command");
+
+    String::from_utf8(sha.stdout).unwrap()
+}
+
 
 fn main() {
     let out_dir = PathBuf::from(env::var("OUT_DIR").unwrap());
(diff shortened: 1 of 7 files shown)


build.rs
```diff
@@ -16,8 +17,11 @@ fn main() {
             .expect("Error during SCSS compilation")
     ).expect("Couldn't write CSS output");
 
+    let cache_id = plume_common::utils::random_hex();
```
igalic: i'd rather this wasn't random, but deterministic

such as the file's she256sum

fdb-hiroshima: What is your issue with it being random? It's fixed at compile time, so assuming a production server, which may restart but won't recompile often, the cache will be ok. My first through was using git commit sha, but this may be unnecessary information disclosure

igalic: if nothing about the code changes, the cache_id shouldn't change either

BaptisteGelez: The build won't be run again if the code didn't changed, so the ID will not change.

build.rs
```diff
@@ -1,8 +1,39 @@
-extern crate plume_common;
```
BaptisteGelez: You should probably remove it from `Cargo.toml` too.

fdb-hiroshima: I totally forgot about that

build.rs
igalic: yes, in our development environment that is true, but not for anyone building packages or container images of our software. they always start from a clean slate

build.rs
```diff
@@ -2,6 +2,38 @@ extern crate ructe;
...
+fn compute_static_hash() -> String {
+    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat --printf='%n %Y\n' | sha256sum"
+
+    let find = Command::new("find")
+        .args(&["static/", "-type", "f", "!", "-path", "static/media/*"])
+        .stdout(Stdio::piped())
+        .spawn()
+        .expect("failed find command");
+
+    let sort = Command::new("sort")
```
igalic: btw, is there no way to do this in rust?

fdb-hiroshima: I think there are plenty, but I'm lazy and it's a build script. I prototyped using bash, so my rust code is basically bash translated (if you find it really bad, I can change that but I don't know if it's really useful)

igalic: nah, i was just curious

src/template_utils.rs
```diff
@@ -9,13 +15,36 @@ pub static CACHE_NAME: &str = env!("CACHE_ID");
...
 pub type BaseContext<'a> = &'a(&'a Connection, &'a Catalog, Option<User>);
 
-pub type Ructe = Content<Vec<u8>>;
+#[derive(Debug)]
+pub struct Ructe(pub Vec<u8>);
+
+impl<'r> Responder<'r> for Ructe {
+    fn respond_to(self, r: &Request) -> response::Result<'r> {
+        //if method is not Get or page contain a form, no caching
+        if r.method() != Method::Get || self.0.windows(6).any(|w| w == b"<form ") {
```
BaptisteGelez: Probably obvious, but why do we disable caching on pages with a form?

fdb-hiroshima: If the page contain a form, the csrf token will be cached, and may no longer match user's cookies. As these tokens are added later by a fairing, there is no way to detect the page changed here. The other solution could be to allow caching of page and for rocket_csrf to remove ETag when appropriate

PR #482, opened 2019-03-16
Add markdown support for summary

### LINKED ISSUE #247: User summary has no markdown support, but allows html
when entering a summary to `http://0.0.0.0:7878/@/:user/edit` that Summary cannot be Markdown.

However, it will accept html… which seems to be normalized:

```
Homoiconic ops witch<br/>Disappointing expectations since 1894.
```

becomes:

```html
<div class="user-summary">
    Homoiconic Ops Witch<br>Disappointing Expectations since 1894.
</div>
```

fdb-hiroshima:
I should probably add a new field to keep raw md
Fixes #247

### DIFF
--- a/migrations/postgres/2019-03-16-143637_summary-md/down.sql
+++ b/migrations/postgres/2019-03-16-143637_summary-md/down.sql
@@ -1,0 +1,2 @@
+-- This file should undo anything in `up.sql`
+ALTER TABLE users DROP COLUMN summary_html;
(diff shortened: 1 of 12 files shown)


BaptisteGelez:
Also once #454 is merged, it would be interesting to use the "inline" mode for Markdown that was introduced, to avoid having profiles with big images or titles in their summaries.

igalic:
idea for testing migrations: run the setup, then run `redo`

fdb-hiroshima:
I don't know how to do that in Travis, but it's actually what I'm doing on my computer

Issue #254, opened 2018-10-01
Hide CWed images


BaptisteGelez:
We can add a CW when uploading a media, but it is never shown anywhere.

The best solution until we have a visual editor is to replace image tags by `summary`/`details` with pulldown-cmark before saving the article.

Plume version (and/or commit): 0.2.0

fdb-hiroshima:
fixed by #483

PR #483, opened 2019-03-16
Hide cw pictures behind a summary/details

fdb-hiroshima:
This is an attempt at fixing #254 
Sadly it doesn't work with the new editor, as html tags are put in a `<code>` bloc


BaptisteGelez:
It makes me notice that one can copy almost any HTML in the editor, which is a bit weird, and full of bug (and maybe of security risks?). I think we should add a listener for `onpaste` and filter what is pasted (to only allow images or simple text)

fdb-hiroshima:
from what I see, any html pasted is surrounded with `<code>`. There are no more risks than with the previous editor. Filtering/modifying what is pasted might be nice in some situations, sometime one just want to put some html elements not supported by plain markdown

fdb-hiroshima:
tested with #486, the md tag is not interpreted as it is inside some html. Maybe I should put plain html instead of markdown when using cw

BaptisteGelez:
I think Markdown in HTML never worked. At least it doesn't on GitHub:

<div>
**hello** I'm in a `div`
</div>

So yes, going with HTML for everything seems to be a better solution. Ideally, we should have some kind of custom Markdown like ![132] where 132 is the idea of a media, and then Plume just replaces it with the full HTML code, but this is probably quite hard to achieve.

fdb-hiroshima:
It might not be that hard actually, we "just" have to match with the Event `Start(Image(_, _, _))` and map it to our custom html

fdb-hiroshima:
User should input something like `some title`
If the corresponding picture has a cw, it display this
Screenshot_2019-04-06_16-40-55
And after clicking on it, it shows the picture (this is a toggle, so clicking again hide it back).

fdb-hiroshima:
Hum, there is a small bug if the same picture with cw is present twice, clicking on the second will only activate the first one. I shall probably add some random to the generated id

fdb-hiroshima:
it's not from your tests. It seems to came from the rule adding a 3em margin on `main article img`

PR #486, opened 2019-03-16
Editor improvements

BaptisteGelez:
- Make it possible to insert new paragraphs in the article body
- Make it impossible to copy formatted HTML (to make media insertion from markdown code work correctly)

TODO:

- [x] make it possible to escape draft mode
- [x] display errors from the server
- [x] button to go back to the "normal" editor
- [x] Avoid publishing placeholders


fdb-hiroshima:
If I may add to the todo list: https://github.com/Plume-org/Plume/pull/458#issuecomment-473323102

BaptisteGelez:
@fdb-hiroshima 	Yes, the first one is already fixed, I'll see for the other. :)

plume-front/src/editor.rs
```diff
@@ -80,10 +80,30 @@ fn init_widget(
...
+            document.execCommand("insertText", false, evt.clipboardData.getData("text"));
+        });
+    };
+}
+
 pub fn init() -> Result<(), EditorError> {
+    // Check if the user wants to use the basic editor
+    if let Some(x) = window().session_storage().get("basic-editor") {
+        if x == String::from("true") {
+            window().session_storage().remove("basic-editor");
```
fdb-hiroshima: I think basic-editor should be permanent, and there would then be a button to re-enable the "advanced" editor

plume-front/src/editor.rs
```diff
@@ -80,10 +80,46 @@ fn init_widget(
...
+        if basic_editor == String::from("true") {
+            if let Some(editor) = document().get_element_by_id("plume-fallback-editor") {
+                if let Ok(Some(title_label)) = document().query_selector("label[for=title]") {
+                    let editor_button = document().create_element("a")?;
+                    js!{ @{&editor_button}.href = "#"; }
+                    editor_button.add_event_listener(|_: ClickEvent| {
+                        window().local_storage().remove("basic-editor");
+                        window().history().go(0).ok(); // refresh
+                    });
+                    editor_button.append_child(&document().create_text_node(&i18n!(CATALOG, "Open the fancy editor")));
```
BaptisteGelez: I don't really know how to name it…

fdb-hiroshima: rich text editor?

BaptisteGelez:
@fdb-hiroshima This should definitely be fixed before merging this branch I think. I will see what I can do.

fdb-hiroshima:
The only fix I see is, given the following :
```html
<div>some text</div><div>![desc](path/to/picture.png)</div>
```
Where we are given a single `Html("<div>some text</div><div>desc</div>")`, we could cut it out as `["some text", "desc"]`, call cmark on each of these, and put back the `div`s

BaptisteGelez:
I was more thinking of pre-processing the HTML with the DOM APIs in WASM, because your solution may break footnotes, or links that use this syntax:

```
my [link]

[link]: https://example.org
```

PR #494, opened 2019-03-20
Centralize configuration and add some new config

fdb-hiroshima:
Ideally, if someone could review the idea in [this comment](https://github.com/Plume-org/Plume/issues/273#issuecomment-474982184), I'd like to add the implementation to this pr too

### DIFF
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -142,7 +142,7 @@ impl Blog {
     }
 
     fn fetch_from_webfinger(conn: &Connection, acct: &str) -> Result<Blog> {
-        resolve(acct.to_owned(), CONFIG.use_https)?
+        resolve(acct.to_owned(), true)?
             .links
             .into_iter()
             .find(|l| l.mime_type == Some(String::from("application/activity+json")))
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -26,7 +26,7 @@ use safe_string::SafeString;
 use schema::blogs;
 use search::Searcher;
 use users::User;
-use {Connection, Error, Result, BASE_URL, USE_HTTPS};
+use {Connection, Error, Result, CONFIG};
 
 pub type CustomGroup = CustomObject<ApSignature, Group>;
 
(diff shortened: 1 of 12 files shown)


plume-models/src/blogs.rs
```diff
@@ -142,7 +142,7 @@ impl Blog {
     }
 
     fn fetch_from_webfinger(conn: &Connection, acct: &str) -> Result<Blog> {
-        resolve(acct.to_owned(), *USE_HTTPS)?
+        resolve(acct.to_owned(), CONFIG.use_https)?
```
BaptisteGelez: Not really related, but I'm thinking about it because I'm seeing it: shouldn't we drop this setting and force people to set up HTTPS. I introduced it to make it easier to develop locally, but now this requires HTTPS too because of CSRF.

src/main.rs
```diff
@@ -116,29 +128,12 @@ Then try to restart Plume.
...
-    let activity_size = &env::var("ACTIVITY_SIZE")
-        .unwrap_or_else(|_| "1024".to_owned())
-        .parse::<u64>()
-        .unwrap();
-    config.set_limits(
-        Limits::new()
-            .limit("forms", form_size * 1024)
-            .limit("json", activity_size * 1024),
-    );
+    config.set_address(CONFIG.rocket.address.clone()).unwrap();
```
BaptisteGelez: Isn't there a way to use the `Config` struct from rocket directly in our `config` module instead of `RocketConfig`?

fdb-hiroshima: `RocketConfig` could become a `Result<rocket::Config>` (Result to denote when there was an error creating the Config, or when there is no secret key). I through of it at some point, but did not remember what I wanted to do when I was about to commit

plume-models/src/blogs.rs
fdb-hiroshima: We should probably, http does not work well anyway

fdb-hiroshima: As someone on Matrix said they were using Plume on a Tor Hidden Service, I tried to see if this won't cause any CSRF issues, and apparently Tor Browser consider hidden service to be secure even without https, that's nice

fdb-hiroshima:
It was incomplete >_<

BaptisteGelez:
Sorry… I didn't saw you added the "Incomplete" label… :woman_facepalming:

fdb-hiroshima:
No worries, I made another pr with what I wanted to add

fdb-hiroshima:
@khimaros you can open an issue with what you think would be a better comportment :wink:

PR #508, opened 2019-03-31
Correct some grammar in the english text

marek-lach:
Hope I didn't forget anything :-)

 - Also corrects some translations, where they're translations of a different source string than needed

Now resolved by #510


BaptisteGelez:
Wow, thank you so much for all that. I just need to check it (but I trust you, it should be good) and to update the templates too before merging.

marek-lach:
> I'll try to fix the build too, because it looks like it is failing.

Hm... yes it does, I'll try to check where the problem could be.

po/plume/plume.pot
```diff
@@ -37,7 +37,7 @@ msgid "{0}'s avatar"
 msgstr ""
 
 # src/routes/blogs.rs:70
-msgid "You need to be logged in order to create a new blog"
+msgid "To create a new blog, you need to be logged in"
```
BaptisteGelez: I thought these messages have already been corrected, but looks like they haven't… :thinking:

BaptisteGelez:
> Edit: There's a difference of one line in pl.pot, as seen bellow:
> 
> Screenshot 2019-03-31 at 20 41 18
> 
> I didn't think that matters, but maybe it does, who knows?

No, I don't think it changes anything.

Also for next time you want to update a lot of messages like that: edit them directly in the `templates` files, I can then copy your branch a generate all the 
`.po`/`.pot` automatically.

po/plume/plume.pot
marek-lach: I did correct it in Crowndin some time back, but it's better to do it at the source level... :)

marek-lach:
> Also for next time you want to update a lot of messages like that: edit them directly in the `templates` files, I can then copy your branch a generate all the
> `.po`/`.pot` automatically.

I have now done this. If possible, you can exclude my .po files from this merge, and rebuild them anew based on the string changes I made in templates.

PR #511, opened 2019-04-01
use circleci instead of travis

fdb-hiroshima:
Upgrade rust-toolchain
Upgrade Tantivy
Warning: tantivy now use a different file format, search index may need to be deleted and recreated to work as intended
try to run compile twice in a row because sometime oomk is evil


.circleci/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
fdb-hiroshima: oops I shouldn't have committed this one :grimacing:

fdb-hiroshima:
Wow the ci is so much faster, it was 15m without with no cache, and now it's under 6. I did not expect it to be that much

PR #513, opened 2019-04-03
Add tests for plume webserver

fdb-hiroshima:
Tests are done using Firefox (I might try to use Chrome too), [Selenium](https://www.seleniumhq.org/projects/webdriver/) and Python (there is no good crate to use Selenium on Rust, so it seemed like a reasonable choice).
Also add a feature `test` for routes that will be necessary for some tests, but should not be on a public server (like committing Tantivy's index on demand instead of every 30 minutes to make possible to test search page)


.circleci/config.yml
```diff
@@ -26,12 +42,14 @@ aliases:
   - &env_postgresql
     environment:
       MIGRATION_DIR: migrations/postgres
+      MIGRATION_DIRECTORY: migrations/postgres
```
BaptisteGelez: Isn't there one of the two lines above that should be removed?

.circleci/images/plume-buildenv/Dockerfile
```diff
@@ -4,16 +4,29 @@ ENV PATH="/root/.cargo/bin:${PATH}"
...
+    cargo install diesel_cli --no-default-features --features postgres,sqlite --version '=1.3.0' &&\
     rm -fr ~/.cargo/registry
 
 #install coverage tools
 RUN curl -L https://github.com/SimonKagstrom/kcov/archive/master.tar.gz | tar xz &&\
     mkdir -p kcov-master/build && cd kcov-master/build && cmake .. && make &&\
     make install && cd ../.. && rm -rf kcov-master
+
+#set some compilation parametters
+RUN echo "W3RhcmdldC54ODZfNjQtdW5rbm93bi1saW51eC1nbnVdCnJ1c3RmbGFncyA9IFsiLUNsaW5rLWRlYWQtY29kZSIsICItQ2xpbmstYXJncz0tWGxpbmtlciAtLW5vLWtlZXAtbWVtb3J5IC1YbGlua2VyIC0tcmVkdWNlLW1lbW9yeS1vdmVyaGVhZHMiXQoK" | base64 -d >> ~/.cargo/config
```
BaptisteGelez: Maybe you could add a comment telling what these parameters are, to make it easier to edit them in the future.

.circleci/config.yml
fdb-hiroshima: gosh my config is so old I had MIGRATION_DIR in it (so I copied it), but I was missing the MIGRATION_DIRECTORY for diesel so I added it, but in fact MIGRATION_DIR is no longer used by us

.circleci/images/plume-buildenv/Dockerfile
fdb-hiroshima: I think I'll change this to a copy statement, it will make more sense

script/run_browser_test.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -eo pipefail
+
+export ROCKET_SECRET_KEY="AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
+
+mkdir -p "target/cov/plume"
+mkdir -p "target/cov/plm"
+plm='kcov --exclude-pattern=/.cargo,/usr/lib --verify target/cov/plm plm'
```
BaptisteGelez: I'm not sure to understand what you are doing here… :thinking:

fdb-hiroshima: I'm creating a "shortcut" to plm run with kcov because I run it multiple time in the following lines, and it's rather long. So when I call `$plm <something>` after, it just mean "run `plm <something>` with coverage report"

BaptisteGelez: Oh OK, I get it, it is to get coverage for plm too. Great!

PR #514, opened 2019-04-03
Corrected grammar pointed out by translators

marek-lach:
The user 'UniqueActive' pointed these issues out on Crowdin.

Thank you!

### DIFF
--- a/templates/instance/about.rs.html
+++ b/templates/instance/about.rs.html
@@ -13,7 +13,7 @@
     <div class="banner">
         <section class="stats">
         <div>
-            <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> users"; n_users))</p>
+            <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> people"; n_users))</p>
         </div>
         <div>
             <p>@Html(i18n!(ctx.1, "Who wrote <em>{0}</em> articles"; n_articles))</p>
--- a/templates/partials/instance_description.rs.html
+++ b/templates/partials/instance_description.rs.html
@@ -21,7 +21,7 @@
                 @Html(instance.short_description_html)
                 <section class="stats">
                     <div>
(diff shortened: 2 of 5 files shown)


templates/partials/instance_description.rs.html
```diff
@@ -21,7 +21,7 @@ <h2>@i18n!(ctx.1, "About {0}"; instance.name)</h2>
                 @Html(instance.short_description_html)
                 <section class="stats">
                     <div>
-                        <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> people"; n_users))</p>
+                        <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> users"; n_users))</p>
```
BaptisteGelez: I would have preferred to use "people" for both, it feels more like they are real human beings than "users" IMO.

marek-lach: Right. Done :-)

marek-lach:
> Also they saw a `'` that shouldn't be here in templates/users/followed.rs.html IIRC, could you please remove it?

Yeah, I cannot for the life of me find where in that file the incorrect ' symbol actually is... hmmm...

templates/instance/about.rs.html
```diff
@@ -13,7 +13,7 @@ <h1>@i18n!(ctx.1, "About {0}"; instance.name)</h1>
     <div class="banner">
         <section class="stats">
         <div>
-            <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> users"; n_users))</p>
+            <p>@Html(i18n!(ctx.1, "Home to <em>{0}</em> people"; n_users))</p>
```
igalic: 👍

would be easier to remember if `n_users` was `n_people`

marek-lach: It's honestly not a big problem. I think from a programmer's perspective, having the declaration as `users` is easier to understand in code maybe?

igalic: true, true…

PR #519, opened 2019-04-08
Add support for remote interact

fdb-hiroshima:
Add support for remote interaction ([this thing](https://eldritch.cafe/users/Barmaid/remote_follow) in mastodon)

- [x] create the endpoint dispatching remote interactions to local pages
- [x] add this endpoint to web-finger
- [x] propose remote interaction when following and not connected
- [x] propose remote interaction when liking/sharing and not connected
- [x] propose remote interaction when commenting and not connected
- [x] fetch posts/comments we don't know but remote interaction was requested for ?


templates/users/header.rs.html
```diff
@@ -32,7 +32,7 @@ <h1 class="grow flex vertical">
             <a class="u-url" href="@user.ap_url"></a>
         }
 
-        @if ctx.2.clone().map(|u| u.id != user.id).unwrap_or(false) {
+        @if ctx.2.clone().map(|u| u.id != user.id).unwrap_or(true) {
```
fdb-hiroshima: (rational behind this : show the button when not connected)

Cargo.toml
```diff
@@ -24,6 +24,7 @@ rocket = "0.4.0"
 rocket_contrib = { version = "0.4.0", features = ["json"] }
 rocket_i18n = "0.4.0"
 rpassword = "2.0"
+runtime-fmt = "0.3.0"
```
igalic: what's this do?

fdb-hiroshima: like format!() or println!(), but with the base string unknown at compile time (when we fetch the url, we get something like `https://<instance>/authorize_interaction?uri={uri}`, which we obviously can't no at compile time)

fdb-hiroshima:
blocked by #443

PR #520, opened 2019-04-08
Standard smallest screen is 600px

marek-lach:
For devices that have a screen size so small they have to use the `max-width` property in CSS, the standard of a screen-size that has to be used here is `phones 600px and smaller`, not `900px`, which would have to use the `min-width` property instead, that would mean a widescreen size, because `min-width` means `a screen size of 900px, or more`, including desktops.

### DIFF
--- a/static/css/_header.scss
+++ b/static/css/_header.scss
@@ -88,7 +88,7 @@ header {
 }
 
 // Only enable label animations on normal screens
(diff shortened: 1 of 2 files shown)


marek-lach:
Woops, missed that. Fixed now. Thanks for reviewing!

PR #528, opened 2019-04-18
try to build release plume&plm on ci

fdb-hiroshima:
ref Plume-org/docs#27
I should archive the artifacts before uploading

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -193,6 +193,7 @@ aliases:
       - run: script/generate_artifact.sh
       - store_artifacts:
           path: plume.tar.gz
+          destination: plume.tar.gz
 
 
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,6 +90,21 @@ aliases:
         - ~/.cargo/
         - target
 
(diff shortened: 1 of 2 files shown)


BaptisteGelez:
And also the migration files?

fdb-hiroshima:
Do I include both migrations, only the one of interest as `migration/$db_name` or only the one of interest as `migration/` (maybe not the latest as it will not be compatible with our documentation?)

igalic:
I'm not sure i've asked this before, but, how hard would it be to support both databases? at the same time?

fdb-hiroshima:
I see 2 ways of doing so : 
 - using `Box<Connection>` everywhere
 - creating an enum whose variants are the different backend we support, which implement Connection (and some others Diesel traits) by calling the function on it's inner type
But I don't know how any of those would play with r2d2 (the connection manager)

BaptisteGelez:
Couldn't it be possible to make all our models generic over the connection type that is used? (but that's probably more work than what you proposed)

And for the migrations, the best is to have only the relevant one in `migrations/$db` to make the archive smaller and keep compatibility with our current docs.

PR #479, opened 2019-03-15
New Crowdin translations

BaptisteGelez:
Enjoy!

Issue #504, opened 2019-03-26
Mobile feed view is too narrow


Granitosaurus:
Feed views on mobile are very narrow which makes for a very difficult browsing. The "latest articles" is the only one not affected by this.

Tested on plume.mastodon.host instance. 

![Narrow](https://i.imgur.com/p3J5dwK.jpg)

BaptisteGelez:
Hello. Do you have any idea what your screen resolution is? And I guess it was on plume.mastodon.**host**, right? Thank you. :slightly_smiling_face:

Granitosaurus:
My phone says 2240 x 1080
Comparison screenshot:
![narrow](https://i.imgur.com/IcAo8Ky.jpg)

marek-lach:
The view shrinks whenever you click on either 'Federated feed', or 'Local feed' on fediverse.blog

As per testing, this issue is **NOT present** on the _cafe.sunbeam.city_ instance.

BaptisteGelez:
I think there is a CSS class that is applied on the article lists on the homepage, but not on the other page, and that is causing this bug.

PR #529, opened 2019-04-19
Version bump: 0.3.0 (alpha 2)

BaptisteGelez:
Needs #479 to be merged first


marek-lach:
And #530 should be merged also, before the official release of Alpha 2 is put out.

Great work @BaptisteGelez, @fdb-hiroshima, @igalic et al!

marek-lach:
Maybe taking a look at #504 is also an option...

PR #533, opened 2019-04-19
strip binary before packing

fdb-hiroshima:
reduce by 35-40% size of binaries and archive


igalic:
does this still allow any kind of sensible debugging?

fdb-hiroshima:
panics and log message are unchanged, but running under gdb or valgrind would be terrible because function names (and other things) are no longer present. I expect someone wanting to use gdb to not use a binary compiled with --release anyway, as it already remove a lot of information that help debugging with such tools

igalic:
thank you very much for the explanation!

PR #540, opened 2019-04-21
Remove Canapi

BaptisteGelez:
It added more complexity than it helped.

I will eventually rewrite it as a proc-macro that generates types for an API from an OpenAPI spec, so maybe it will be re-introduced at some point (but it will be totally different).

Also I added an endpoint to delete posts, the function was here before, but not exposed.


plume-api/src/posts.rs
```diff
@@ -1,13 +1,10 @@
...
-    pub title: Option<String>,
+pub struct NewPostData {
+    pub title: String,
     pub subtitle: Option<String>,
-    pub content: Option<String>,
-    pub source: Option<String>,
-    pub author: Option<String>,
+    pub source: String,
+    pub author: String,
+    // If None, a blog for this author will be choosen automatically
```
fdb-hiroshima: I think it should be chosen automatically only if the user has only one blog, so to be clearly deterministic (currently it's choosing the first found, which was already the old behavior)

PR #545, opened 2019-04-24
Make comments in styleshhets a bit clearer

marek-lach:


static/css/_variables.scss
```diff
@@ -1,4 +1,4 @@
-// Colors
+// Color Scheme
```
igalic: how did this change, in its own, fail the build?

fdb-hiroshima: kcov seems to use something that is not always available in circle-ci containers, and make the job fail with a "set personality failed" or similar. When I see one, I usually restart the workflow to make things pass as they should, in this case I wasn't quick enough and you noticed before me

PR #546, opened 2019-04-25
Make submit forms/buttons more responsive for mobile

marek-lach:
A blind attempt to resolve submit buttons not resizing on very small screens and input forms not resizing either.

This would need to be tested on a test setup first, which I didn't test, to figure out if it didn't break any desktop sizes.


fdb-hiroshima:
You can test it live on https://pr-546.joinplu.me/

PR #547, opened 2019-04-25
Remove unnecessary webkit styling line

marek-lach:


static/css/_forms.scss
```diff
@@ -54,7 +54,6 @@ form.inline {
     cursor: pointer;
     font-size: 1em;
     width: auto;
-   -webkit-appearance: none;
```
marek-lach: The intent here was just to remove the `-webkit-appearance: none;` on line 57, because it does not seem to be necessary for all of the buttons to work correctly anyway without this one line. It therefore has no actuall purpose and essentially does nothing in practice, I think. I could be wrong on this, as I haven't tested pressing the submit comments button, but as far as I can tell the line should be removed in the interest of cleaner code.

BaptisteGelez: Could you please check, with an iOS device (as I think you have one, right?), on https://pr-547.joinplu.me/ that no button look like the one on the left of this picture?

image

I think this line was added to prevent this kind of styling issue.

But if you already tested and everything looks fine, I can trust you and merge this PR. :slightly_smiling_face:

(I have no iOS device to check by myself, sorry…)

marek-lach:
No, no, then it serves it purpose as it should :-) Closing now...

PR #549, opened 2019-04-25
[WIP] Fix menu not unfolding on iOS

marek-lach:


marek-lach:
> If you go on https://pr-549.joinplu.me with a small screen (or by reducing the size of your browser window) you will see that the elements of the menu are always visible and overflow on the right.
> 
> I think there was a script to do that on browser that don't support :focus-within (like Safari or Chrome), but maybe it is not working? Which version of iOS are you using (if you were the one who found this bug)?

You are right. Honestly, I am not 100% sure how to adress this now, but the menu works fine in Plume Alpha 2, so the change occured between then and now. I'll keep experimenting further and see if I can figure something out... if not, I will close this after some time and think about it later.

BaptisteGelez:
Can you see the new editor on iOS? I'm asking to be sure the WASM scripts are correctly run.

marek-lach:
> Can you see the new editor on iOS? I'm asking to be sure the WASM scripts are correctly run.

Just tested it on the newest iOS. Yes, the editor itself works fine, it's just that the menu doesn't unfold. I think some styling with `webkit-appearence:...` around the element that handles that may fix it, but am not sure exactly. On Android all is fine as always :-)

PR #550, opened 2019-04-27
Restore menu.js file for iOS

marek-lach:


BaptisteGelez:
Before testing, this script should be loaded, so you should probably add it to `templates/base.rs.html` too.

marek-lach:
Yep, just testing.

marek-lach:
Well, the best thing would be to probably compare the old and the new script for the menu, and see where a mistake occured.

Closing this for now.

PR #551, opened 2019-04-27
Federate user deletion

### LINKED ISSUE #509: User deletion is not federated
When deleting an account on an instance, it is still visible on the others.

- **Plume version:** 0.2.0

BaptisteGelez:
- When someone deletes their account
- When a local user is banned

Fixes #509

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1777,7 +1777,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "plume"
-version = "0.2.0"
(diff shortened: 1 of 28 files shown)


Cargo.lock
```diff
@@ -1777,7 +1777,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "plume"
-version = "0.2.0"
+version = "0.3.0"
```
BaptisteGelez: I'm not sure why this is here?

PR #552, opened 2019-04-27
Add some feedback when performing some actions

### LINKED ISSUE #468: More feedback when doing an action
When deleting something, or publishing an article for instance, a little message should appear at top of the page (just under the header) to inform you everything went well.

BaptisteGelez:
Here is an example of what it looks like (this is just a test, in reality you can only have one of these messages at the same time):

Screenshot_2019-04-27 Plume unu ⋅ Plume

I don't know if there is a simpler way to do it. Also, I added `msg: Option<FlashMessage>` to almost all routes to be sure that if a message was present it would be rendered, but maybe it is not necessary and only make the code harder to read? And maybe `msg` should go in `PlumeRocket` as it is used almost everywhere?

Fixes #468

### DIFF
--- a/po/plume/hr.po
+++ b/po/plume/hr.po
@@ -49,6 +49,14 @@ msgstr ""
 msgid "A blog with the same name already exists."
 msgstr ""
 
+# src/routes/blogs.rs:142
+msgid "Your blog was successfully created!"
+msgstr ""
+
+# src/routes/blogs.rs:167
+msgid "Your blog was deleted."
+msgstr ""
+
 # src/routes/blogs.rs:172
 msgid "You are not allowed to delete this blog."
 msgstr ""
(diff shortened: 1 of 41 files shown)


fdb-hiroshima:
> And maybe `msg` should go in `PlumeRocket` as it is used almost everywhere?

Considering how it must be everywhere, I think it can go in `PlumeRocket`, as well as in template `BaseContext`. Also `BaseContext` could implement `From<PlumeRocket>` so it is easier to build and reduce duplication

po/plume/hr.po
```diff
@@ -65,10 +73,56 @@ msgstr ""
...
 msgstr ""
 
+# src/routes/medias.rs:143
+msgid "Your media have been deleted."
+msgstr ""
+
+# src/routes/posts.rs:140
+#, fuzzy
+msgid "You are not allowed to delete this media."
+msgstr "Ti ne autor ovog bloga."
```
igalic: this is supposed to be

> Ti nisi autor ovog(a?) bloga

but i just realised that fuzzy means it was auto translated…

BaptisteGelez:
@igalic I don't really know… it is true it pollutes a little bit the diff, but at the same time it makes sense to have them with the change in the interface.

po/plume/hr.po
BaptisteGelez: Yes, it was auto-translated. Once it will be on Crowdin, you will be able to fix it. :wink:

fdb-hiroshima:
I'd be to not pollute pr with this, and have something that autocommit the changes once a week, or after each merge

src/template_utils.rs
```diff
@@ -13,7 +13,41 @@ pub use askama_escape::escape;
...
 
-pub type BaseContext<'a> = &'a (&'a Connection, &'a Catalog, Option<User>);
+pub type BaseContext<'a> = &'a (
+    &'a Connection,
+    &'a Catalog,
+    Option<User>,
+    Option<(String, String)>,
+);
+
+pub trait IntoContext {
```
fdb-hiroshima: why not implementing `From<PlumeRocket>` for Context? It will auto implement `Into<Context>` for PlumeRocket, and you should be able to use `into()` with no need to import this trait everywhere

BaptisteGelez: Because Rust is complaining about the fact that all the parameters of the trait comes from another crate, and that it could cause conflicts, while I don't think it can? :woman_shrugging: But yeah, I tried it, and it doesn't work.

fdb-hiroshima: ho, it's because BaseContext is not a true struct but a typedef to a tuple of basic types (in theory it could cause conflict, because the exact same trait can be implemented in plume-model, thus creating a conflict). This mean we can either convert BaseContext to be a true type, or implement Into in plume-models. I've tried the first one, but it requires some small but ugly changes so I'm not sure it would be better (you can take a look at the patch [here](https://framadrop.org/r/aXcBuKkLEW#rzbPsNSdq+XCFboTnRLdY8Qt3RNqqUYGgrnA8bfUoQQ=))

BaptisteGelez:
Do you want me to apply your patch before merging?

fdb-hiroshima:
It was just to show what I came up with, but having `&(&rockets).into()` everywhere is definitely worse than importing a trait

PR #553, opened 2019-04-27
Theme update

BaptisteGelez:
So far:

- Ligther colors
- No more border radius
- Buttons are now always colored
- Fix weird margin with CWed images
- Add a border to quotes
- Start to redesign the post page (according to the Figma mockups)
  - Full-size cover image, with a gradient and the title/subtitle/authors/date over it
  - Like/boost are now between tags and author info
  - Author info has a difference background color to make a difference between the different sections
  - Tags and license are now side-by-side

As I started it, this redesign is not perfectly following what was done on Figma (for instance lighter colors were not planned), but I'm open to discussion if you think Figma mockups should be better respected.

It is available at https://pr-553.joinplu.me/ if you want to see how it looks like. You can login with admin/admin123 to see all the pages.


trwnh:
Perhaps `main article`'s max-width should be a bit wider? `40rem` is quite narrow. Optimal readability would suggest making it somewhere between 45-75 characters, which can be represented by the `ch` unit in CSS. So I would propose perhaps making the rule `max-width: 70ch` instead of `max-width: 40rem`

trwnh:
Also the `main header` and `main header > img` are way too tall, and should each have a max-height of no more than `calc(100vh - $heightOfBodyHeader)` -- this allows the header to not be bigger than the browser viewport.

BaptisteGelez:
So… I tried to limit the height of the illustration, but it is impossible to do it without losing the ratio or a part of the image. I don't know what should be done: hide some a portion of the illustration if it is too big, or give the responsibility to author to provide images in reasonable dimensions?

Someone also proposed to have an option that could be enabled to limit the height of the header (by hiding some of the image) if you want to be sure to never have giant images. So that can be a solution too.

trwnh:
I think fitting is better than cropping. You can either accept the ratio loss inherent to cover-fit, or you can cap the max-width of the image/header (in addition to the max-height). That style of header image can be seen in Ghost, for example: https://demo.ghost.io/welcome/

marek-lach:
This is a great change. I only propose changing the main purple color of links from the current hex code of `#7765E3` to something like `#887ADE`, `#8D80DE`, or perhaps `#8F85D5`, because on very light backgrounds, it's very hard to see, as shown bellow:
Screenshot 2019-04-30 at 19 46 41

Screenshot 2019-04-30 at 19 53 18

BaptisteGelez:
@marek-lach Even with another shade of purple the contrast was quite poor, so I made it white, but with underlining when you pass your cursor over it to make you understand it is a link.

static/css/_global.scss
```diff
@@ -297,13 +326,37 @@ p.error {
 body > footer {
   display: flex;
   align-content: center;
```
marek-lach: Under that we could also add the line:
`display: -webkit-flex;` to ensure Safari likes it :-)

BaptisteGelez: According to [this table](https://developer.mozilla.org/en-US/docs/Web/CSS/display#Browser_compatibility) Safari supports `display: flex` even without a specific prefix.

marek-lach:
Maybe under the `About this instance` link in the footer, there could also be a `Privacy policy` link. It would help to fill up the empty space there a little bit.

As per what the `Privacy policy page` itself could contain, just basic stuff could be something like:

If you are browsing this site as **a visitor**, your IP address is collected by the server for moderation purposes.
As a `registered user`, you also have to provide your username (which does not have to be your real name), your functional email address and a password, in order to be able to log in, write articles and comment. The content you submit is stored until you delete it.

BaptisteGelez:
I added it. I just changed a bit your text, because we don't actually collect IP addresses, and I added a part about which cookies we are using.

marek-lach:
> I added it. I just changed a bit your text, because we don't actually collect IP addresses, and I added a part about which cookies we are using.

Sure 👍

fdb-hiroshima:
I think this should be a default privacy policy only, and should be editable easily by admins (probably in a future pr, this one is supposed to be about design). While _we_ don't store much (no ip, no cookies when not logged in...), most reverse proxy configuration will log at least user ip.

static/css/_variables.scss
```diff
@@ -1,16 +1,22 @@
 // Color Scheme
-$background: #F4F4F4;
-$form-input-background: #FFFFFF;
-$gray: #E3E3E3;
-$lightgray: #ECECEC;
+$gray: #F3F3F3;
 $black: #242424;
-$white: #F4F4F4;
+$white: #FDFDFD;
```
fdb-hiroshima: this value might be a bit too close to actual white : pages that are mainly forms (creating article, editing profile, or even logging in) look odd in my opinion. I've tried with `F8F8F8`, and I think it looks better so to contrast with `FFFFFF` inside input elements

static/css/_global.scss
```diff
@@ -125,7 +133,7 @@ main {
 p.error {
  	color: $red;
  	font-weight: bold;
- 	max-width: 40rem;
+ 	max-width: 40rem; // TODO: change this value
```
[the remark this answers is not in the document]
BaptisteGelez: I think I introduced a variable for that value, but maybe I didn't?

PR #555, opened 2019-04-28
import migrations and don't require diesel_cli for admins

fdb-hiroshima:
this is an attempt at integrating migrations to plm, with the goal to remove need for diesel_cli, and to allow arbitrary rust in migrations (can come handy for many reasons, like need to upgrade tantivy index version due to upstream changes, without need to take special steps for sysadmins)

- [x] backbone for the feature
- [x] integration to plm as new sub-command
- [x] update ci
- [x] make plume fail to launch if it detect migrations must be run

### DIFF
--- a/plume-macro/src/lib.rs
+++ b/plume-macro/src/lib.rs
@@ -85,10 +85,14 @@ fn file_to_migration(file: &str) -> TokenStream2 {
     for line in file.lines() {
         if sql {
             if line.starts_with("--#!") {
-                actions.push(quote!(Action::Sql(#acc)));
+                if !acc.trim().is_empty() {
+                    actions.push(quote!(Action::Sql(#acc)));
+                }
                 sql = false;
                 acc = line[4..].to_string();
                 acc.push('\n');
+            } else if line.starts_with("--") {
+                continue;
             } else {
                 acc.push_str(line);
                 acc.push('\n');
@@ -97,6 +101,8 @@ fn file_to_migration(file: &str) -> TokenStream2 {
             if line.starts_with("--#!") {
                 acc.push_str(&line[4..]);
                 acc.push('\n');
+            } else if line.starts_with("--") {
+                continue;
             } else {
                 let func: TokenStream2 = trampoline(TokenStream::from_str(&acc).unwrap().into());
                 actions.push(quote!(Action::Function(&#func)));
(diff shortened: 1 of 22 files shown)


script/run_unit_test.sh
```diff
@@ -4,6 +4,10 @@ for file in target/debug/*-*[^\.d]; do
 	if [[ -x "$file" ]]
 	then
 		filename=$(basename $file)
+		if [[ "plume_macro-c53c304b7b109570" =~ ^plume_macro ]]; then
```
BaptisteGelez: Isn't this always true? (maybe not, I'm really bad with Bash)

fdb-hiroshima: oops I tested with a value, and forgot to use the variable after :grimacing:

plume-macro/src/lib.rs
```diff
@@ -0,0 +1,139 @@
...
+                .read_to_string(&mut down_sql)
+                .unwrap();
+            let name = path
+                .file_name()
+                .unwrap()
+                .to_str()
+                .unwrap()
+                .chars()
+                .filter(char::is_ascii_digit)
+                .take(14)
```
igalic: this seems like a strange thing to `take()`
what is this whole expression supposed to do?

fdb-hiroshima: diesel store whether migrations were run by storing a 14 digit code. For `2019-03-06-115158_blog_images`, it is translated to `20190306115158`. To do that I get an iterator over the folder name, filter only digits, and take the first 14 ones (if there are more, they are ignored)

plume-models/src/migrations.rs
```diff
@@ -53,7 +53,7 @@ impl ImportedMigrations {
         let latest_migration = conn.latest_run_migration_version()?;
         let latest_id = if let Some(migration) = latest_migration {
             self.0
-                .binary_search_by_key(&migration.as_str(), |mig| mig.name)
+                .binary_search_by_key(&dbg!(migration.as_str()), |mig| dbg!(mig.name))
```
BaptisteGelez: Should these `dbg!` be here?

fdb-hiroshima: no :x

fdb-hiroshima:
At some point the documentation should also be updated to reflect that change

fdb-hiroshima:
@BaptisteGelez I think it's possible, I'll try to do it (diesel_cli does that from what I saw in it's code)

igalic:
> I don't think it would be possible, but creating the PostgreSQL database if it doesn't exist would be perfect.

usually, create db is a more privileged operation than creating anything in that db…
and i sure hope people aren't giving our software that kind of permissions 😅

fdb-hiroshima:
So, shall I try, or do I assume people have a "secure" configuration that don't allow us to do privileged things?

BaptisteGelez:
Let's say people have a secure configuration. :smile:

PR #556, opened 2019-04-29
remove the new index lock tantivy uses

fdb-hiroshima:


fulmicoton:
Sorry for chiming in a bit late. In recent version, tantivy uses bsd locks. You don't need to remove those files at all. If your process die, restarting should work just fine.

fdb-hiroshima:
I've noticed it while trying my changes, but actually one user managed to have it's index locked while no plume was running, I've no idea how. Anyway it's here in case it happen, and hopefully it does not

PR #557, opened 2019-04-30
reduce reqwest timeout to 5s

fdb-hiroshima:
Current timeout is 30s, this add a second kind of timeout : if connection to remote can't be established in 5s, the remote is considered down, so that a thread don't stall for too long for no reason (and also make some unit test much faster to run)


fdb-hiroshima:
maybe not a unique client (if we have multiple request to do at the same time they will be queued), but we could probably make some kind of pool of clients. I'll keep the idea somewhere for when I have spare time to spend

PR #558, opened 2019-05-01
rewrite circleci config

fdb-hiroshima:
- [x] use better syntax
- [x] grcov for coverage
- [x] automatize crowdin interactions


.circleci/images/plume-buildenv/cargo_config
```diff
@@ -1,3 +1,3 @@
 [target.x86_64-unknown-linux-gnu]
 # link dead code for coverage, attempt to reduce linking memory usage to not get killed
-rustflags = ["-Clink-dead-code", "-Clink-args=-Xlinker --no-keep-memory -Xlinker --reduce-memory-overheads"]
+rustflags = ["-Clink-args=-Xlinker --no-keep-memory -Xlinker --reduce-memory-overheads"]
```
BaptisteGelez: Don't we need to link dead code anymore?

fdb-hiroshima: It's somewhere it this line now https://github.com/Plume-org/Plume/pull/558/files#diff-1d37e48f9ceff6d8030570cd36286a61R74

.circleci/config.yml
```diff
@@ -1,314 +1,232 @@
...
-      - release_postgresql:
-          requires:
-            - all_ok
-      - release_sqlite:
-          requires:
-            - all_ok
+    - cargo fmt
+    - clippy:
+        postgres: false
+    - clippy:
```
BaptisteGelez: Could it be possible to give more distinct name to these? I think that because of that we get one job called "clippy" (or whatever) and another one "cliipy-1", but it is not clear which one uses SQlite or PostgreSQL.

fdb-hiroshima: not without having to duplicate config sadly :/

BaptisteGelez: OK, let's forget about it then, it was more bonus.

fdb-hiroshima: kinda related, I'd prefer to use an enum telling we should use sqlite or postgres, but there is no way to do conditional based on other things than boolean :/
And there is no way to select if we use a docker or not, just to select which one, so I run empty alpine when nothing is needed.
I feel like this new syntax is better than the previous, but it misses some core features to not feel hacky in many places

.circleci/config.yml
```diff
@@ -96,9 +96,11 @@ commands:
         default: false
     steps:
     - run: |
-       cargo build <<#parameters.release>>--release<</parameters.release>> --no-default-features --features="${FEATURES}" -p <<parameters.package>> ||
-       cargo build <<#parameters.release>>--release<</parameters.release>> --no-default-features --features="${FEATURES}" -p <<parameters.package>>
-
+            cmd="cargo build <<#parameters.release>>--release<</parameters.release>> --no-default-features --features="${FEATURES}" -p <<parameters.package>>"
+            for i in {1..16}; do
+                $cmd && exit 0
```
igalic: that seems excessive

BaptisteGelez: Yeah, but don't worry, this was an old version, now we only repeat it a few times:

```bash
for i in 36 4 2 1 1; do
   ...
```

where `i` is the number of threads of the command.

It is to avoid job failures because of OOM errors.

fdb-hiroshima: I think you're reviewing an older commit

PR #560, opened 2019-05-03
Fix certain improper rendering of forms

marek-lach:
Attempt at fixing the jumbled styling of the search form in Apple's Safari, which is not present in Blink browsers like Opera, or Chrome:
Screenshot 2019-05-03 at 21 28 28

It appears I forgot about that one in my PR some months back which addressed how forms in Safari were displayed (_maybe search wasn't even there then...?_).

Safari does seem a bit like the new Internet Explorer these days, unfortunately.

This is otherwise a rather small and safe PR for merging, since it has practically no reason to break anything anywhere else.


BaptisteGelez:
I don't think this issue has anything to do with `display: flex`. Safari should support it normally. I think it is more that it tries to apply a "native" style to search fields by default ([`<input type="search">`](https://github.com/Plume-org/Plume/blob/master/templates/search/index.rs.html#L9)), and that a CSS rule to prevent that should be added [here](https://github.com/Plume-org/Plume/blob/master/static/css/_forms.scss#L6):

```css
appearence: none;
```

But I tested it, and it doesn't seem to break anything with Firefox, so it is still safe to merge if you want.

BaptisteGelez:
I also merged master into this branch, so it should be available at https://pr-560.joinplu.me soon, if you want to test in Safari.

marek-lach:
> I don't think this issue has anything to do with `display: flex`. Safari should support it normally. I think it is more that it tries to apply a "native" style to search fields by default ([`<input type="search">`](https://github.com/Plume-org/Plume/blob/master/templates/search/index.rs.html#L9)), and that a CSS rule to prevent that should be added [here](https://github.com/Plume-org/Plume/blob/master/static/css/_forms.scss#L6):
> 
> ```css
> appearence: none;
> ```
I actually do think you are correct, but just in case that didn't sufficiently fix the issue with the search bar (or any other input/submit form), I didn't want to do 2 separate PRs with basically the same objective in the near future, so I basically just put everything in relation to Safari into this one PR and hopefully that would be sufficient for anything Apple may decide to break with how Safari displays things for the forseeable future :-)

templates/search/index.rs.html
```diff
@@ -7,6 +7,7 @@
   <h1>@i18n!(ctx.1, "Search")</h1>
   <form method="get" id="form">
     <input id="q" name="q" placeholder="@i18n!(ctx.1, "Your query")" type="search">
+      <style type="text/css">-webkit-appearance: none;</style>
```
BaptisteGelez: Why don't you add it to the stylesheet directly?

marek-lach: Where do you mean, I am not too sure...sorry. In the `.css` files I have adressed every form and `input` field that I felt like it needed to be adressed. I have reformatted the line you higlighted, so it should be all fine now, I guess? :-)

marek-lach:
The appearence of the search bar is now properly fixed:

Screenshot 2019-05-05 at 12 25 48

templates/search/index.rs.html
BaptisteGelez: (sorry for the late reply)

You actually forgot this one: https://github.com/Plume-org/Plume/pull/560/files#diff-339bac14a232fb01418ef93c2cb5e1ecR6

I think if you add `appearence: none` here, you could remove all the inline styles and have them all in one place.

marek-lach:
@BaptisteGelez This PR is ready to be merged, I think, once the build tests suceed.

static/css/_forms.scss
```diff
@@ -96,7 +100,10 @@ form.inline {
     }
  	}
 }
-input[type="submit"] { display: block; }
+input[type="submit"] {
+ display: block;
+ -webkit-appearance: none;
```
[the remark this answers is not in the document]
marek-lach: Right. As you say, it's best to make these changes to be as universal, and platform agnostic as possible in the long-term. 
I have now ammended the commits.

templates/search/index.rs.html
marek-lach: Unfortunately, in some cases, like with imput forms and buttons, there just has to be the -webkit prefix present, otherwise the buttons are disorted when it's just appearence: none, and so the desired effect is not achievable otherwise it seems like...

marek-lach:
@BaptisteGelez So I finally tested the compiled version on the live instance both in desktop Safari for OSX and in mobile Safari for iOS and it looks nice!

Screenshot 2019-05-06 at 18 54 51

Even things like native shortcuts (notice the little `x` at the end) work in the searchbar now, which was not the case before. That is honestly an even better result than I had originally hoped for.

I think this PR is now really ready :-)

By the way, I tested in Firefox too and everything is there just as it was before, so nothing's broken:
Screenshot 2019-05-06 at 19 07 10

PR #561, opened 2019-05-04
Improve notification management

### LINKED ISSUE #472: Make sure we don't save notifications for remote users
I think we currently do.

### LINKED ISSUE #505: Notify for either comments or mentions
#499 (re)introduced notifications for comments but now you get notified twice, if mentioned in a comment.

Receive a notification for comments on own posts but only if you are not also mentioned in the comment. If mentioned only add notification about the mention.

**Additional context**
#137 already discussed this issue but thanks to me, the problem appeared again :)

fdb-hiroshima:
fix #472 
fix #505

### DIFF
--- a/plume-models/src/comments.rs
+++ b/plume-models/src/comments.rs
@@ -157,17 +157,19 @@ impl Comment {
 
     pub fn notify(&self, conn: &Connection) -> Result<()> {
         for author in self.get_post(conn)?.get_authors(conn)? {
-            if Mention::list_for_comment(conn, self.id)?.iter().all(|m| m.get_mentioned(conn).map(|u| u != author).unwrap_or(true)) {
-                if author.is_local() {
-                    Notification::insert(
-                        conn,
-                        NewNotification {
-                            kind: notification_kind::COMMENT.to_string(),
-                            object_id: self.id,
-                            user_id: author.id,
-                        },
-                    )?;
-                }
+            if Mention::list_for_comment(conn, self.id)?
+                .iter()
+                .all(|m| m.get_mentioned(conn).map(|u| u != author).unwrap_or(true))
+                && author.is_local()
+            {
+                Notification::insert(
+                    conn,
+                    NewNotification {
+                        kind: notification_kind::COMMENT.to_string(),
+                        object_id: self.id,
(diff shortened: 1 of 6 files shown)


plume-models/src/comments.rs
```diff
@@ -157,14 +157,20 @@ impl Comment {
...
-                conn,
-                NewNotification {
-                    kind: notification_kind::COMMENT.to_string(),
-                    object_id: self.id,
-                    user_id: author.id,
-                },
-            )?;
+            if Mention::list_for_comment(conn, self.id)?
+                .iter()
+                .all(|m| m.get_mentioned(conn).map(|u| u != author).unwrap_or(true))
```
BaptisteGelez: Shouln't this be `.any(...)`? Because I'm still getting duplicated notifications.

plume-models/src/comments.rs
```diff
@@ -157,14 +157,20 @@ impl Comment {
...
+                && author.is_local()
```
BaptisteGelez: Also, `AsActor::is_local` is technically broken, since it only does `instance_id == 1` which is not always correct (even if it is most of the time (like, unless you ran `plm instance new` twice, it will probably be correct)).

Maybe we should have a global storing the local instance info somewhere BTW (loaded only once when the app starts)? It would probably avoid a lot of queries to the database, and would allow to fix this issue without changing `AsActor` definition (but probably in another PR if we ever do it).

plume-models/src/comments.rs
fdb-hiroshima: if *all* mentioned users are different from this author, send notification. I think it's correct, and when I tested it, it seemed to work

BaptisteGelez: Oh, I found what is wrong I think: the mention was inserted after the comment in the database.

src/routes/comments.rs
```diff
@@ -80,6 +79,8 @@ pub fn create(
                 .expect("comments::create: mention save error");
             }
 
+            comm.notify(&*conn).expect("comments::create: notify error");
```
BaptisteGelez: OK, I'm too late, you already found it.

PR #568, opened 2019-05-07
use openssl instead of sha256sum for build.rs

### LINKED ISSUE #566: Build error: sha256sum not found on FreeBSD
```
error: failed to compile `plume v0.3.0 (/usr/local/src/Plume)`, intermediate artifacts can be found at `/usr/local/src/Plume/target`                                         
                                                                                                                                                                              
Caused by:                                                                                                                                                                   
  failed to run custom build command for `plume v0.3.0 (/usr/local/src/Plume)`                                                                                               
process didn't exit successfully: `/usr/local/src/Plume/target/release/build/plume-e42edb5ba3415496/build-script-build` (exit code: 101)                                     
--- stdout                                                                                                                                                                   
cargo:rerun-if-changed=/usr/local/src/Plume/templates                                                                                                                        
cargo:rerun-if-changed=/usr/local/src/Plume/templates/blogs                                                                                                                  
…etc…
cargo:rerun-if-changed=/usr/local/src/Plume/templates/partials/instance_description.rs.html                                                                                  
cargo:rerun-if-changed=/usr/local/src/Plume/templates/base.rs.html                                                                                                           
cargo:rerun-if-changed=static/css                                                                                                                                            
                                                                                                                                                                             
--- stderr                                                                                                                                                                   
thread 'main' panicked at 'failed sha256sum command: Os { code: 2, kind: NotFound, message: "No such file or directory" }', src/libcore/result.rs:999:5stat: illegal option --
 -                                                                                                                                                                           
                                                                                                                                                                             
usage: stat [-FLnq] [-f format | -l | -r | -s | -x] [-t timefmt] [file|handle ...]                                                                                           
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.                                                                                               
                                                                                                                                                                             
```

Since our build already depends on openssl-dev, consider using `openssl dgst sha256sum`.

fdb-hiroshima:
fix #566

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -5,7 +5,7 @@ use std::process::{Command, Stdio};
 use std::{env, fs::*, io::Write, path::PathBuf};
 
 fn compute_static_hash() -> String {
-    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat --printf='%n %Y\n' | openssl dgst -r"
+    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat -c'%n %Y' | openssl dgst -r"
 
     let find = Command::new("find")
         .args(&["static/", "-type", "f", "!", "-path", "static/media/*"])
@@ -20,7 +20,7 @@ fn compute_static_hash() -> String {
         .expect("failed sort command");
 
     let xargs = Command::new("xargs")
-        .args(&["stat", "--printf='%n %Y\n'"])
+        .args(&["stat", "-c'%n %Y'"])
         .stdin(sort.stdout.unwrap())
         .stdout(Stdio::piped())
         .spawn()
--- a/build.rs
+++ b/build.rs
@@ -5,7 +5,7 @@ use std::process::{Command, Stdio};
 use std::{env, fs::*, io::Write, path::PathBuf};
 
 fn compute_static_hash() -> String {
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -5,7 +5,7 @@ use std::process::{Command, Stdio};
 use std::{env, fs::*, io::Write, path::PathBuf};
 
 fn compute_static_hash() -> String {
-    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat --printf='%n %Y\n' | sha256sum"
+    //"find static/ -type f ! -path 'static/media/*' | sort | xargs stat --printf='%n %Y\n' | openssl dgst -r"
```
igalic: i'd prefer we leave `stat(1)` out of this, as the `--printf` option is not platform independent, and no equivalent, platform independent option exists

the only other would be to use `-c`

fdb-hiroshima: The goal of this function is to have a unique id, and best effort is made to not invalidate cache if not necessary. We can also not make that best effort (recompiling Plume => new cache id) and just generate a random hex string. That would be platform independent.
What shall I do?

igalic: i think i'd just go with `stat -c` for now

there's not a big chance we'll be building on windows… easily… i think, so let's just go with this.

PR #571, opened 2019-05-10
upload artifact of wasm binary

### LINKED ISSUE #570: Release `static` folder separate from bin & migrations
compiling wasm requires rust-lld, which is only enabled (and compiles) on Tier 1 platforms.

Given that the produced wasm & js are platform independent, we should release them as separate download artefacts.

fdb-hiroshima:
fix #570

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -216,6 +216,9 @@ jobs:
     - store_artifacts:
         path: plume.tar.gz
         destination: plume.tar.gz
+    - store_artifacts:
+        path: wasm.tar.gz
+        destination: wasm.tar.gz
 
   push translations:
     executor:
--- a/script/generate_artifact.sh
+++ b/script/generate_artifact.sh
@@ -2,4 +2,5 @@
 mkdir bin
 cp target/release/{plume,plm} bin
 strip -s bin/*
-tar -cvzf plume.tar.gz bin/ static/ migrations/$FEATURES
+tar -cvzf plume.tar.gz bin/ static/
+tar -cvzf wasm.tar.gz static/plume-front.{js,wasm}


script/generate_artifact.sh
```diff
@@ -2,4 +2,5 @@
 mkdir bin
 cp target/release/{plume,plm} bin
 strip -s bin/*
-tar -cvzf plume.tar.gz bin/ static/ migrations/$FEATURES
+tar -cvzf plume.tar.gz bin/ static/
```
igalic: this one should be with `migrations/$FEATURES`, and without `static/`

script/generate_artifact.sh
```diff
@@ -2,4 +2,5 @@
...
+tar -cvzf wasm.tar.gz static/plume-front.{js,wasm}
```
igalic: This one, should contain all of `static/` — it's entirely architecture independent!

.circleci/config.yml
```diff
@@ -216,6 +216,9 @@ jobs:
     - store_artifacts:
         path: plume.tar.gz
         destination: plume.tar.gz
+    - store_artifacts:
+        path: wasm.tar.gz
+        destination: wasm.tar.gz
```
igalic: and as such, I'd just call it static.tar.gz

(should this have a version number)

script/generate_artifact.sh
fdb-hiroshima: migrations don't require to be shipped as they are integrated into `plm`, and I think it should be with `static/` so that for x86_64 users, there is only a single archive to download, un-tgz, and run

script/generate_artifact.sh
fdb-hiroshima: I expected it to be an alternative step to all `cargo web` things in [here](https://docs.joinplu.me/installation/with/source-code), in which case it doesn't require the whole `static/` dir as it is `git clone`d. What were your expectations?

igalic: you are right.

PR #572, opened 2019-05-10
cache local instance

### LINKED ISSUE #564: cache local instance
cache local instance to reduce useless query made at the database (see https://github.com/Plume-org/Plume/pull/561#discussion_r280983340)

fdb-hiroshima:
fix #564

### DIFF
--- a/plume-cli/src/main.rs
+++ b/plume-cli/src/main.rs
@@ -6,7 +6,7 @@ extern crate rpassword;
 
 use clap::App;
 use diesel::Connection;
-use plume_models::{Connection as Conn, CONFIG, instance::Instance};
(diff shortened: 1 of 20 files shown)


plume-cli/src/main.rs
```diff
@@ -27,6 +27,7 @@ fn main() {
 
     dotenv::dotenv().ok();
     let conn = Conn::establish(CONFIG.database_url.as_str());
+    let _ = conn.as_ref().map(|conn| Instance::cache_local(conn));
```
BaptisteGelez: Do you really need this `let` here?

fdb-hiroshima: just to explicitly ignore a warning about unused result

Issue #567, opened 2019-05-06
Compatability with AP-enabled Wordpress blogs


DeadSuperHero:
Hey there, Plume team!

I'm currently trying out your software on a public instance (fediverse.blog), and was interested in the possibility of subscribing to my ActivityPub-enabled Wordpress blog, which is made possible by [Pterotype](https://github.com/pterotype-project/pterotype). For context, I run a site called [WeDistribute.org](https://wedistribute.org), which talks about fediverse projects. Currently, a few types of platforms are capable of subscribing to it:

* Mastodon
* Pleroma
* Friendica
* Hubzilla

I think, in the long term, it could be very crucial for ActivityPub-enabled blogging platforms to be capable of federating with one another. Currently, I am unable to follow my Wordpress blog from Plume. Attempting  to navigate to ether `https://fediverse.blog/@/blog@wedistribute.org` or `https://fediverse.blog/~/blog@wedistribute.org` does not appear to bear fruit yet.

![image](https://user-images.githubusercontent.com/401560/57260434-68b92180-7018-11e9-81c1-389c5fec9edb.png)

My question to you is, what would it take to make cross-federation with Plume possible? How might separate ActivityPub-enabled article platforms work to ensure compatability with each other?

fdb-hiroshima:
I'm looking into it. It seems the webfinger entry for the user `blog@wedistribute.org` contain escaped `/`, where we expect theme to be used directly (it's how other implementations behave). I'm not sure if it should be considered a Plume bug, or a Pterotype one, but being a bit more liberal on this can't really harm

marek-lach:
This issue, related to Plume's URL svhema retreival may be of some help in sheding light on the issue in general, even if it's not specifically Wordpress related:

https://github.com/tsileo/microblog.pub/issues/49

fdb-hiroshima:
The two issues are different and unrelated, but I'll try to push fixes for both by the end of the day

marek-lach:
> The two issues are different and unrelated, but I'll try to push fixes for both by the end of the day

Oh... good to know, sorry :-)

pfefferle:
Hey I am the developer of the ActivityPub plugin for WordPress. I just tested my implementation with plume and it seems to detect the profile, but my blog does not receive a follow request.

PR #573, opened 2019-05-11
Fix some federation issue

fdb-hiroshima:
The issue in #567 will be a bit tougher than initially expected. Pterotype seems to use Organization instead of plain User for federation, so there is the question how do we manage that? How do we treat that? As users I expect, but it will require some changes to do so, or as something else?


BaptisteGelez:
A post can't have an associated blog, but it technically can have no author. So I would say Organizations should be treated as blogs.

You can probably create a type that implements `AsActor` and the different `AsObject`, but that actually delegates the work to `Blog`? I don't know.

fdb-hiroshima:
atm a post can't really have no authors : in multiple places assumption is made that there is at least one author, and we can use `posts.get_authors()?[0]` safely. There will be things to change if we introduce the possibility of a post without authors. Also as one can't follow a blog yet, this would mean a basic feature is lacking. The more I think about it, the more problematic it get

BaptisteGelez:
OK, so then, what about inserting both a blog and an user with the same username when encountering an `Organization`? It is a bit hackish, but it should work.

Also, we should maybe add a "default" personnal blog for users at some, as I think it is what most people expect (but it may make the whole code a lot more complex too)?

marek-lach:
For reference:

https://github.com/w3c/activitypub/issues/260

marek-lach:
@fdb-hiroshima @BaptisteGelez According to the creator of _Pterotype_, `Organization` is the entire blog, not a specific user in the current design of the plugin.

I think this may help to shed some light on the issue at hand:
Screenshot 2019-05-16 at 17 12 46

fdb-hiroshima:
this has been stall for too long. What has been fixed has been fixed, the rest will be at some point in the future

AnaGelez:
Wait I just realized: do we fetch blogs with the `group:` prefix in Webfinger now?

fdb-hiroshima:
```sh
$ curl 'https://pr-573.joinplu.me/.well-known/webfinger?resource=group:Test@pr-573.joinplu.me'
{
   "subject":"acct:Test@pr-573.joinplu.me",
   "aliases":[
      "https://pr-573.joinplu.me/~/Test/"
   ],
   "links":[
      {
         "rel":"http://webfinger.net/rel/profile-page",
         "href":"https://pr-573.joinplu.me/~/Test/"
      },
      {
         "rel":"http://schemas.google.com/g/2010#updates-from",
         "href":"https://pr-573.joinplu.me/~/Test/feed.atom",
         "type":"application/atom+xml"
      },
      {
         "rel":"self",
         "href":"https://pr-573.joinplu.me/~/Test/",
         "type":"application/activity+json"
      }
   ]
}
```

fdb-hiroshima:
no, we don't 😬

PR #580, opened 2019-05-14
Mobile menu fix for webkit

marek-lach:
This should fix the mobile menu in portrait mode not unfolding on iOS.

- [ ] Need to test it on a live device

### DIFF
--- a/plume-front/src/main.rs
+++ b/plume-front/src/main.rs
@@ -68,16 +68,16 @@ fn main() {
 /// It should normally be working fine even without this code
 /// But :focus-within is not yet supported by Webkit
 fn menu() {
-    if let Some(button) = document().get_element_by_id("menu" onclick="") {
+    if let Some(button) = document().get_element_by_id("menu") {
         if let Some(menu) = document().get_element_by_id("content") {
             button.add_event_listener(|_: ClickEvent| {
                 document()
-                    .get_element_by_id("menu" onclick="")
+                    .get_element_by_id("menu")
                     .map(|menu| menu.class_list().add("show"));
             });
             menu.add_event_listener(|_: ClickEvent| {
                 document()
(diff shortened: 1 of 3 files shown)


plume-front/src/main.rs
```diff
@@ -66,18 +66,18 @@ fn main() {
 /// Toggle menu on mobile device
 ///
 /// It should normally be working fine even without this code
-/// But :focus-within is not yet supported by Webkit/Blink
+/// But :focus-within is not yet supported by Webkit
 fn menu() {
-    if let Some(button) = document().get_element_by_id("menu") {
+    if let Some(button) = document().get_element_by_id("menu" onclick="") {
```
BaptisteGelez: I am not sure what you wanted to do here, but this syntax is not valid. `get_element_by_id` only takes one string parameters, that is the `id` of the element to select.

If you wanted to select `#menu` element with an empty `onclick` property, you can probably use `document().query_selector("#menu[onclick=""]")` (this functions takes a CSS selector). But I'm not even sure it would work, because I don't think using `add_event_listener` like we do changes `onclick`.

marek-lach: Yeah, honestly, I wasn't too sure that would work either. The CSS `cursor: pointer ` attribute should be enough for this. I'll see if that works...

BaptisteGelez:
What is the issue you are trying to fix exactly? (I'm just curious because I can't see any bug with Chrome on Android)

marek-lach:
It does not happen on Android. Or anywhere else really, except in mobile Safari. I'll try this last approach and if it doesn't work, I'll just leave it be.

marek-lach:
A solution similar to Android's simulated click functionality is apparently at least in development by Apple: https://webkit.org/status/#?search=pointer%20events.

PR #596, opened 2019-05-24
feature: custom domains using Fairings

igalic:
This Pull Request addresses #94; replaces #447;
We start by adding `custom_domain` to the `blogs` table; and a custom type Host(String)

A blog with `custom_domain` has special routes with higher priority, but administrative routes will remain on the instance host.

We'll use `Fairings` to extract the `Host` header, to then specially route them.

Domain administration is left to the user (and/or admin)
(but we might be able to provide a check on whether it's already correctly configured)

-------

## TODO

- [x] PostgreSQL Migrations
- [x] SQLite Migrations
- [x] Custom time to represent domain (`Host`)
- [x] Expand `Blog` (and `NewBlog`)
- [x] Update `/blog/new` and `/blog/edit` to add custom domain field
- [x] Cache custom domains on startup and on insert/update.
- [x] `/custom_domains/` mount-point which handles these domains
- [ ] implement routes for
    - [x] blog details
    - [x] post details 
    - [x] blog activity_details
    - [ ] search
    - [ ] routes necessary for federation??
- [ ] implement redirect for the "legacy" routes for the above
    - [x] blog details
    - [x] post details
    - [ ] blog activity_details ??
    - [ ] search
- [x] Update URLs in templates
- [ ] Update URLs in base template for search
- [ ] Check correct domain setup
- [ ] Test: Federation
- [ ] Some explanations (in the UI, next to the setting and in the docs)

### DIFF
--- a/migrations/sqlite/2019-05-22-124153_add_custom_domains/up.sql
+++ b/migrations/sqlite/2019-05-22-124153_add_custom_domains/up.sql
@@ -1,0 +1,57 @@
+-- add custom_domain to blogs
+CREATE TABLE IF NOT EXISTS "blogs_add_custom_domain" (
+  id INTEGER NOT NULL PRIMARY KEY AUTOINCREMENT,
+  actor_id VARCHAR NOT NULL,
+  title VARCHAR NOT NULL,
+  summary TEXT NOT NULL DEFAULT '',
+  outbox_url VARCHAR NOT NULL UNIQUE,
+  inbox_url VARCHAR NOT NULL UNIQUE,
+  instance_id INTEGER REFERENCES instances(id) ON DELETE CASCADE NOT NULL,
+  creation_date DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+  ap_url text not null default '' UNIQUE,
+  private_key TEXT,
+  public_key TEXT NOT NULL DEFAULT '',
+  fqn TEXT NOT NULL DEFAULT '',
+  summary_html TEXT NOT NULL DEFAULT '',
+  icon_id INTEGER REFERENCES medias(id) ON DELETE SET NULL DEFAULT NULL,
+  banner_id INTEGER REFERENCES medias(id) ON DELETE SET NULL DEFAULT NULL,
+  custom_domain text default NULL UNIQUE,
+  CONSTRAINT blog_unique UNIQUE (actor_id, instance_id)
+);
+
+INSERT INTO blogs_add_custom_domain (
+  id,
+  actor_id,
+  title,
+  summary,
+  outbox_url,
+  inbox_url,
+  instance_id,
+  creation_date,
+  ap_url,
+  private_key,
+  public_key,
+  fqn,
+  summary_html,
+  icon_id,
+  banner_id
+) SELECT
+  id,
+  actor_id,
+  title,
+  summary,
+  outbox_url,
+  inbox_url,
+  instance_id,
+  creation_date,
+  ap_url,
+  private_key,
+  public_key,
+  fqn,
+  summary_html,
+  icon_id,
+  banner_id
+FROM blogs;
+
+DROP TABLE blogs;
+ALTER TABLE "blogs_add_custom_domain" RENAME to blogs;
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -335,7 +335,7 @@ impl Blog {
                 .unwrap_or_else(|| Host::new(format!(
                     "{}/~/{}",
                     Instance::get_local().unwrap().public_domain,
-                    self.title
+                    self.fqn,
                 )))
         )
     }
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -81,7 +81,7 @@ pub struct NewBlog {
     pub summary_html: SafeString,
     pub icon_id: Option<i32>,
     pub banner_id: Option<i32>,
-    pub custom_domain: Option<String>,
+    pub custom_domain: Option<Host>,
 }
 
 const BLOG_PREFIX: &str = "~";
@@ -310,6 +310,15 @@ impl Blog {
             .execute(conn)
             .map(|_| ())
             .map_err(Error::from)
+    }
+
+    pub fn list_custom_domains(conn: &Connection) -> Result<Vec<String>> {
+        blogs::table
+            .filter(blogs::custom_domain.is_not_null())
+            .select(blogs::custom_domain)
+            .load::<Option<String>>(conn)
+            .map_err(Error::from)
+            .map(|res| res.into_iter().map(|s| s.unwrap()).collect::<Vec<_>>())
     }
 }
 
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -25,6 +25,7 @@ use posts::Post;
 use safe_string::SafeString;
 use schema::blogs;
 use search::Searcher;
+use std::fmt::{self, Display};
 use users::User;
 use {Connection, Error, PlumeRocket, Result};
 
@@ -36,6 +37,12 @@ pub struct Host(String);
 impl Host {
     pub fn new(host: impl ToString) -> Host {
         Host(host.to_string())
+    }
+}
+
+impl Display for Host {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
     }
 }
 
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -195,6 +195,23 @@ impl Blog {
         }
     }
 
+    pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
+        let from_db = blogs::table
+            .filter(blogs::custom_domain.eq(host))
+            .limit(1)
+            .load::<Blog>(&*c.conn)?
+            .into_iter()
+            .next();
+        if let Some(from_db) = from_db {
+            Ok(from_db)
+        } else {
+            // we should never get here, because
+            // a) load::<Blog>()? should return early if it fails
+            // b) this function is only called after a Request::guard::<Host>()
+            Err(Error::NotFound)
+        }
+    }
+
     fn fetch_from_webfinger(c: &PlumeRocket, acct: &str) -> Result<Blog> {
         resolve_with_prefix(Prefix::Group, acct.to_owned(), true)?
             .links
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -1,6 +1,6 @@
 use activitypub::{actor::Group, collection::OrderedCollection, object::Image, CustomObject};
 use chrono::NaiveDateTime;
-use diesel::{self, ExpressionMethods, QueryDsl, QueryResult, RunQueryDsl, SaveChangesDsl};
+use diesel::{self, ExpressionMethods, QueryDsl, RunQueryDsl, SaveChangesDsl};
 use openssl::{
     hash::MessageDigest,
     pkey::{PKey, Private},
@@ -195,10 +195,11 @@ impl Blog {
         }
     }
 
-    pub fn find_by_host(c: &PlumeRocket, host: Host) -> QueryResult<Blog> {
+    pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
         blogs::table
             .filter(blogs::custom_domain.eq(host))
             .first::<Blog>(&*c.conn)
+            .map_err(|_| Error::NotFound)
     }
 
     fn fetch_from_webfinger(c: &PlumeRocket, acct: &str) -> Result<Blog> {
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -8,8 +8,9 @@ use openssl::{
     sign::{Signer, Verifier},
 };
 use rocket::{
+    http::RawStr,
     outcome::IntoOutcome,
-    request::{self, FromRequest, Request},
+    request::{self, FromFormValue, FromRequest, Request},
 };
 use serde_json;
 use url::Url;
@@ -68,6 +69,15 @@ impl<'a, 'r> FromRequest<'a, 'r> for Host {
                 }
             })
             .or_forward(())
+    }
+}
+
+impl<'v> FromFormValue<'v> for Host {
+    type Error = &'v RawStr;
+
+    fn from_form_value(form_value: &'v RawStr) -> std::result::Result<Host, &'v RawStr> {
+        let val = String::from_form_value(form_value)?;
+        Ok(Host::new(&val))
     }
 }
 
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -44,6 +44,7 @@ pub struct Blog {
     pub summary_html: SafeString,
     pub icon_id: Option<i32>,
     pub banner_id: Option<i32>,
+    pub custom_domain: Option<String>,
 }
 
 #[derive(Default, Insertable)]
@@ -61,6 +62,7 @@ pub struct NewBlog {
     pub summary_html: SafeString,
     pub icon_id: Option<i32>,
     pub banner_id: Option<i32>,
+    pub custom_domain: Option<String>,
 }
 
 const BLOG_PREFIX: &str = "~";
@@ -392,6 +394,7 @@ impl FromId<PlumeRocket> for Blog {
                         .summary_string()
                         .unwrap_or_default(),
                 ),
+                custom_domain: None,
             },
         )
     }
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -7,6 +7,10 @@ use openssl::{
     rsa::Rsa,
     sign::{Signer, Verifier},
 };
+use rocket::{
+    outcome::IntoOutcome,
+    request::{self, FromRequest, Request},
+};
 use serde_json;
 use url::Url;
 use webfinger::*;
@@ -21,10 +25,44 @@ use posts::Post;
 use safe_string::SafeString;
 use schema::blogs;
 use search::Searcher;
+use std::fmt;
 use users::User;
 use {Connection, Error, PlumeRocket, Result};
 
 pub type CustomGroup = CustomObject<ApSignature, Group>;
+
+#[derive(Queryable, Clone)]
+pub struct Host(String);
+
+impl Host {
+    pub fn new<T: Into<String>>(host: T) -> Host {
+        Host(host.into())
+    }
+}
+
+impl Into<String> for Host {
+    fn into(self) -> String {
+        self.0.clone()
+    }
+}
+
+impl From<String> for Host {
+    fn from(s: String) -> Host {
+        Host::new(s)
+    }
+}
+
+impl AsRef<str> for Host {
+    fn as_ref(&self) -> &str {
+        &self.0
+    }
+}
+
+impl std::fmt::Display for Host {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
 
 #[derive(Queryable, Identifiable, Clone, AsChangeset)]
 #[changeset_options(treat_none_as_null = "true")]
@@ -291,6 +329,24 @@ impl Blog {
             .execute(conn)
             .map(|_| ())
             .map_err(Error::from)
+    }
+}
+
+impl<'a, 'r> FromRequest<'a, 'r> for Host {
+    type Error = ();
+
+    fn from_request(request: &'a Request<'r>) -> request::Outcome<Host, ()> {
+        request
+            .headers()
+            .get_one("Host")
+            .and_then(|x| {
+                if x != Instance::get_local().ok()?.public_domain {
+                    Some(Host(x.to_string()))
+                } else {
+                    None
+                }
+            })
+            .or_forward(())
     }
 }
 
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -202,14 +202,7 @@ impl Blog {
             .load::<Blog>(&*c.conn)?
             .into_iter()
             .next();
-        if let Some(from_db) = from_db {
-            Ok(from_db)
-        } else {
-            // we should never get here, because
-            // a) load::<Blog>()? should return early if it fails
-            // b) this function is only called after a Request::guard::<Host>()
-            Err(Error::NotFound)
-        }
+        from_db.ok_or(Error::NotFound)
     }
 
     fn fetch_from_webfinger(c: &PlumeRocket, acct: &str) -> Result<Blog> {
--- a/plume-models/src/blogs.rs
+++ b/plume-models/src/blogs.rs
@@ -1,6 +1,6 @@
 use activitypub::{actor::Group, collection::OrderedCollection, object::Image, CustomObject};
 use chrono::NaiveDateTime;
-use diesel::{self, ExpressionMethods, QueryDsl, RunQueryDsl, SaveChangesDsl};
+use diesel::{self, ExpressionMethods, QueryDsl, QueryResult, RunQueryDsl, SaveChangesDsl};
 use openssl::{
     hash::MessageDigest,
     pkey::{PKey, Private},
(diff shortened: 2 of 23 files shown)


src/main.rs
```diff
@@ -156,6 +156,9 @@ Then try to restart Plume
         println!("Please refer to the documentation to see how to configure it.");
     }
 
+    let custom_domains = plume_models::blogs::Blog::list_custom_domains(&dbpool.get().unwrap()).unwrap();
+    dbg!(custom_domains);
```
fdb-hiroshima: Although it's very useful for debugging and fine for now, I'm gonna point this just to make sure we don't forget it when time for merging come

plume-models/src/blogs.rs
```diff
@@ -311,6 +311,15 @@ impl Blog {
...
             .map_err(Error::from)
     }
+
+    pub fn list_custom_domains(conn: &Connection) -> Result<Vec<String>> {
+        blogs::table
+            .filter(blogs::custom_domain.is_not_null())
+            .select(blogs::custom_domain)
+            .load::<Option<String>>(conn)
+            .map_err(Error::from)
+            .map(|res| res.into_iter().map(|s| s.unwrap()).collect::<Vec<_>>())
```
fdb-hiroshima:
```
error: redundant closure found
   --> plume-models/src/blogs.rs:321:44
    |
321 |             .map(|res| res.into_iter().map(|s| s.unwrap()).collect::<Vec<_>>())
    |                                            ^^^^^^^^^^^^^^ help: remove closure as shown: `std::option::Option::unwrap`
    |
    = note: `-D clippy::redundant-closure` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#redundant_closure
```

plume-models/src/blogs.rs
```diff
@@ -195,6 +195,23 @@ impl Blog {
...
     }
 
+    pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
+        let from_db = blogs::table
+            .filter(blogs::custom_domain.eq(host))
+            .limit(1)
+            .load::<Blog>(&*c.conn)?
+            .into_iter()
+            .next();
+        if let Some(from_db) = from_db {
```
fdb-hiroshima: you might wanna check [ok_or](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.ok_or)

plume-models/src/blogs.rs
```diff
@@ -195,6 +195,23 @@ impl Blog {
...
+            Ok(from_db)
+        } else {
+            // we should never get here, because
```
fdb-hiroshima: (this is actually false, load can return a success with zero elements (type `select * from blogs where custom_domain = 'none';` in an sql shell if you need to convince yourself, no error, but nothing matched))

plume-models/src/blogs.rs
```diff
@@ -195,6 +195,23 @@ impl Blog {
         }
     }
 
+    pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
+        let from_db = blogs::table
+            .filter(blogs::custom_domain.eq(host))
+            .limit(1)
+            .load::<Blog>(&*c.conn)?
+            .into_iter()
+            .next();
```
fdb-hiroshima: I know this not used everywhere else, but using [first](https://docs.diesel.rs/diesel/query_dsl/trait.RunQueryDsl.html#method.first) would be more meaning full

plume-models/src/blogs.rs
igalic: i understand, but this function is only called after our `guard` already put us here

plume-models/src/blogs.rs
igalic: `limit(1)` also doesn't make sense given that `custom_domain` is `UNIQUE`

fdb-hiroshima: agreed. If I'm correct however, first returns a single entry, not a Vec where you have to try to take the first

igalic:
```rust
    pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
        let from_db = blogs::table
            .filter(blogs::custom_domain.eq(host))
            .load::<Blog>(&*c.conn)?
            .first();
        from_db.ok_or(Error::NotFound)
    }
```
results in
```
error[E0308]: mismatched types
   --> plume-models/src/blogs.rs:203:9
    |
198 |     pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
    |                                                         ------------ expected `std::result::Result<blogs::Blog, Error>` because of return type
...
203 |         from_db.ok_or(Error::NotFound)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected struct `blogs::Blog`, found reference
    |
    = note: expected type `std::result::Result<blogs::Blog, _>`
               found type `std::result::Result<&blogs::Blog, _>`

error: aborting due to previous error
```
i don't know if to laugh or cry.

fdb-hiroshima: I meant [diesel first](https://docs.diesel.rs/diesel/query_dsl/trait.RunQueryDsl.html#method.first), in place of load. Not [Vec first](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.first) (actually it's slice's, but Vec deref to slice)

igalic:
```
error[E0308]: mismatched types
   --> plume-models/src/blogs.rs:199:9
    |
198 |       pub fn find_by_host(c: &PlumeRocket, host: Host) -> Result<Blog> {
    |                                                           ------------ expected `std::result::Result<blogs::Blog, Error>` because of return type
199 | /         blogs::table
200 | |             .filter(blogs::custom_domain.eq(host))
201 | |             .first::<Blog>(&*c.conn)
    | |____________________________________^ expected enum `Error`, found enum `diesel::result::Error`
    |
    = note: expected type `std::result::Result<_, Error>`
               found type `std::result::Result<_, diesel::result::Error>`
```

but if i use `QueryResult<Blog>`, then i can't use `?` for error handling later 🙍

plume-models/src/blogs.rs
igalic: removed comment.

plume-models/src/blogs.rs
igalic: actually, `rustc` is saying exactly what i need to do, so i did that.

plume-models/src/blogs.rs
```diff
@@ -195,14 +195,10 @@ impl Blog {
...
-        let from_db = blogs::table
+    pub fn find_by_host(c: &PlumeRocket, host: Host) -> QueryResult<Blog> {
+        blogs::table
             .filter(blogs::custom_domain.eq(host))
-            .limit(1)
-            .load::<Blog>(&*c.conn)?
-            .into_iter()
-            .next();
-        from_db.ok_or(Error::NotFound)
+            .first::<Blog>(&*c.conn)
```
BaptisteGelez: What about adding a `map_err(|_| Error::NotFound)` here, to keep `Result<Blog>` as the return type (with the advantages it has, like having the possibility to "throw" it to get an error page)?

BaptisteGelez: And also it would allow you to remove the `From<diesel::result::Error> for ErrorPage` impl.

fdb-hiroshima: > What about adding a `map_err(|_| Error::NotFound)` here

or `map_err(Error::from)` ?

igalic: seemed like a good deal to me to directly map a `diesel::result::Error` to an `ErrorPage`
otoh, i can also see how this would lead to the kind of error handling that PHP + MySQL are famous for

igalic:
so: I'm a bit at an impasse or crossroads with this patch, and it all kinda depends on how far i wanna push it.

For now, i've added a 404 handler, which redirects to the main-site, which means, actually, everything works now 😹 (by redirecting to the main site)

We could present the complete illusion that this custom domain, is the real domain, and all there is is this one blog.

To do this, I would have to adapt the URLs `ap_url`, `inbox_url` and `outbox_url` on blog creation (or update) and i would have to modify the templates to use the correct URL scheme.
Alternatively we create an URL builder which we resolves the correct URLs.

Currently, there are a lot of places in which URLs are built ad-hoc.
In some cases, this is okay, because it's just pointing to a rocket route, but in many cases it feels painful, because we're doing manual labour that a function could remove.

the `compute_box()` function for instance, feels quite pointless. There is no computation happening at all, just concatenation! `compute_box()` itself calls `ap_url()` which only prepends `https://`

I would like to propose a Universal URL builder… or rather, a plumey-versal URL builder, which can be used from any part of plume's code to construct a correct plume URL.

fdb-hiroshima:
> To do this, I would have to adapt the URLs ap_url, inbox_url and outbox_url on blog creation (or update) and i would have to modify the templates to use the correct URL scheme.

I think it would be best to still handle most of (~all) ap stuff on the main domain, having an id (as requested in #296) for the activity that goes to the main domain, with a redirect to the correct path when queried (so either the main domain with better name than the id, or custom one when applicable). Also with some webfinger trick it should be possible to have more than one handle for a blog, so there could be the one from the main domain, that shouldn't change because it would break a lot of federation stuff, and an alias based on the custom domain. All this would be in the spirit of not breaking things if the custom domain for a blog changes

> I would like to propose a Universal URL builder… or rather, a plumey-versal URL builder, which can be used from any part of plume's code to construct a correct plume URL.

that would be a good idea for everything that need to integrate the domain in addition to a path. It could be a small macro that call the rocket one, and prepand the domain name, either from local_instance if none is given, or the one given (which would be a custom domain)

igalic:
okay, so, at this point, i'm about as done as i can be, without touching the templates or refactoring the URL building.

migrations/sqlite/2019-05-22-124153_add_custom_domains/up.sql
```diff
@@ -0,0 +1,57 @@
+-- add custom_domain to blogs
+CREATE TABLE IF NOT EXISTS "blogs_add_custom_domain" (
```
rfwatson: Just wondering why copying the table is required, instead of just adding the column?

igalic: because [SQLite supports a limited subset of ALTER TABLE](https://www.sqlite.org/lang_altertable.html)

rfwatson: Damn, I missed it was SQLite. Thanks.

src/routes/blogs.rs
```diff
@@ -35,7 +37,63 @@ pub fn details(name: String, page: Option<Page>, rockets: PlumeRocket) -> Result
...
+    // check this first, and return early
+    // doing this prevents partially moving `blog` into the `match (tuple)`,
+    // which makes it impossible to reuse then.
+    if blog.custom_domain == None {
+        return detail_guts(blog, page, rockets);
+    }
+
+    match (blog.custom_domain, page) {
+        (Some(ref custom_domain), Some(ref page)) => {
+            Ok(Redirect::to(format!("https://{}/?{}", custom_domain, page)).into())
```
BaptisteGelez: I think it will not properly format the page: you will get something like `https://doma.in/?2` when you probably want `?page=2`.

elegaanz (commented):
I gave it a try, and it seems to work quite well so far! I left a few comment about the code itself.

src/routes/posts.rs
```diff
@@ -31,28 +31,14 @@ use routes::{
 };
 use template_utils::{IntoContext, Ructe};
 
-#[get("/~/<blog>/<slug>?<responding_to>", rank = 4)]
-pub fn details(
-    blog: String,
-    slug: String,
+fn detail_guts(
+    blog: &Blog,
+    post: &Post,
```
BaptisteGelez: Does that (and the `blog: &Blog` above) work?

BaptisteGelez: nvm, I thought it was a route, but it is just a regular function.

src/routes/posts.rs
```diff
@@ -94,7 +80,79 @@ pub fn details(
...
+    // check this first, and return early
+    // doing this prevents partially moving `blog` into the `match (tuple)`,
+    // which makes it impossible to reuse then.
+    if blog.custom_domain == None {
+        return detail_guts(&blog, &post, responding_to, &rockets);
+    }
+
+    match (blog.custom_domain, responding_to) {
+        (Some(ref custom_domain), Some(ref responding_to)) => Ok(Redirect::to(format!(
+            "https://{}/?responding_to={}",
```
BaptisteGelez: I think this URL (and the one after) should include the article's slug.

src/routes/posts.rs
```diff
@@ -94,7 +80,79 @@ pub fn details(
...
+            custom_domain, responding_to
+        ))
+        .into()),
+        (Some(ref custom_domain), _) => {
+            Ok(Redirect::to(format!("https://{}/", custom_domain)).into())
+        }
+        (None, _) => panic!("This code path should have already been handled!"),
```
fdb-hiroshima: You can use the handy `unreachable!("<msg>")` instead of a plain panic, it has the same effect but is probably clearer

src/routes/posts.rs
igalic: oh gosh yes that was a copy paste!

src/routes/blogs.rs
```diff
@@ -90,6 +155,12 @@ pub fn create(form: LenientForm<NewBlogForm>, rockets: PlumeRocket) -> RespondOr
     let intl = &rockets.intl.catalog;
     let user = rockets.user.clone().unwrap();
 
+    let custom_domain = if *(&form.custom_domain.is_empty()) {
```
fdb-hiroshima: clippy says 
```suggestion
    let custom_domain = if form.custom_domain.is_empty() {
```

AnaGelez: I can't accept your proposal via Github (apparently I don't have the right to push to this repository :upside_down_face:), but I made it manually, thanks.

src/routes/blogs.rs
```diff
@@ -72,6 +136,7 @@ pub fn new_auth(i18n: I18n) -> Flash<Redirect> {
 pub struct NewBlogForm {
     #[validate(custom(function = "valid_slug", message = "Invalid name"))]
     pub title: String,
```
igalic: we can add a validate function here, but it seems rather limited what we can return

for instance: DNS doesn't resolve? we can't say anything at this point, but should still tell the person that it doesn't resolve (yet)

another test would then be to try to connect and see if it's connecting to us, but the question here is: how??

igalic (commented):
validation?

src/routes/blogs.rs
igalic: would https://github.com/Plume-org/Plume/issues/296 be a solution to this issue?

AnaGelez: I think we should just make a DNS request to check that there is a CNAME record pointing to the main domain? And if it doesn't resolve, we can probably show a warning like: "Your custom domain is not properly configured yet, please re-try later in your blog settings." (or something like that, I trust you to find a better wording).

And I don't think #296 is a solution to this specific issue, even if it may help somewhere else in this branch…

fdb-hiroshima: In some complex setup, checking only a CNAME can be not enough, for instance at home I have `subdomain.my_domain` => `my_domain`, but `my_domain` is also a CNAME to a domain provided by my isp. Suppose Plume run on subdomain.my_domain, pointing to `my_domain`, or A/AAAA to my public ips would work, but would also be rejected. I think having an endpoint which we can query with Reqwest is more solid to complex/unexpected setup

src/routes/blogs.rs
```diff
@@ -136,11 +209,19 @@ pub fn create(form: LenientForm<NewBlogForm>, rockets: PlumeRocket) -> RespondOr
...
-    )
-    .into()
+    if dns_ok {
+        Flash::success(
+            Redirect::to(uri!(details: name = slug.clone(), page = _)),
+            &i18n!(intl, "Your blog was successfully created!"),
+        )
+        .into()
+    } else {
+        Flash::warn(
```
fdb-hiroshima:
```suggestion
        Flash::warning(
```

src/routes/blogs.rs
AnaGelez: And it avoids having to do some DNS queries directly, so even better :smile:

igalic: what would we be looking for when making such a Reqwest?
how do we identify our own instance?

fdb-hiroshima: We could query `/nodeinfo/2.0` and metadata.nodeName matches local value. It does not protect against voluntary "mistake" (actually, nothing can), but will catch configuration error

igalic: is there a cryptographically secure  way in ActivityPub or OcapPub to do this?

fdb-hiroshima: I don't know OcapPub, AP has no cryptography (non normative security sections everywhere), but in general, no cryptographic protection has any effect against a proxy attack, someone could just temporary proxy traffic to Plume, then do nothing else (that's a big problem on smart-cards btw, they tend to use timing info to fight against, but it's not very efficient). The errors we can catch here are mistakes, not clever enough attacks

src/routes/blogs.rs
```diff
@@ -95,6 +95,11 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
     ))
 }
 
+#[get("/validate/<validation_id>")]
+pub fn validate(validation_id: String, rockets: PlumeRocket) -> Result {
```
fdb-hiroshima: the name mismatch the one in main.rs, and Result take 2 type arguments (probably both String I think)

igalic: thanks / sorry for the noise 🙇‍♀️

the two files were edited with about a week or two in between 😅

fdb-hiroshima: np :wink:

src/routes/blogs.rs
```diff
@@ -96,8 +98,27 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
+    if value.is_none() {
+        return Err(String::from("404"));
+    }
+
+    // we have valid id, now check the time
+    let valid_until = value.unwrap();
+    let now = Instant::now();
+
+    // nope, expired (410: gone)
+    if now.checked_duration_since(*valid_until).unwrap().as_secs() > 0 {
```
fdb-hiroshima: you should use `duration_since` instead, as you're unwrapping it it's doing exactly the same thing, and it's stable

src/routes/blogs.rs
```diff
@@ -96,8 +98,27 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
+        valid_domains.inner().remove(validation_id.as_str());
+        return Err(String::from("410"));
```
fdb-hiroshima: (this will only return the string 410, not the error code)

src/routes/blogs.rs
```diff
@@ -96,8 +98,27 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
 #[get("/validate/<validation_id>")]
-pub fn validate(validation_id: String, rockets: PlumeRocket) -> Result {
-    unimplemented!("No idea what to do here yet")
+pub fn domain_validation(
+    validation_id: String,
+    valid_domains: State<HashMap<&str, Instant>>,
+) -> Result<String, String> {
+    let value = valid_domains.inner().get(validation_id.as_str());
+    if value.is_none() {
+        return Err(String::from("404"));
```
fdb-hiroshima: (this will only return the string 404, not the error code)

src/main.rs
```diff
@@ -196,6 +197,7 @@ Then try to restart Plume
         }
     });
 
+    let mut valid_domains: HashMap<&str, Instant> = HashMap::new();
```
fdb-hiroshima: I'd be in favor of using the newtype design pattern for this, so that it make more sense what it is, even at the type level.
I don't know where are coming the &str from, but you might want to use String instead, or you'll have some difficulties with lifetime

trinity-1686a (commented):
:eyes:

src/main.rs
igalic: aye. i haven't gotten that far yet, because i haven't implemented the other side of this

src/routes/blogs.rs
```diff
@@ -100,25 +101,29 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
+    valid_domains: State<Mutex<HashMap<&str, Instant>>>,
+) -> Status {
+    let mutex = valid_domains.inner().lock();
+    let mut validation_map = mutex.unwrap();
+    let validation_getter = validation_map.clone();
+
+    let value = validation_getter.get(validation_id.as_str());
     if value.is_none() {
-        return Err(String::from("404"));
+        return Status::new(404, "validation id not found");
```
fdb-hiroshima: I believe there are consts that already contains standards http status in the module

src/routes/blogs.rs
```diff
@@ -100,25 +101,29 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
 
     // we have valid id, now check the time
     let valid_until = value.unwrap();
     let now = Instant::now();
 
     // nope, expired (410: gone)
-    if now.checked_duration_since(*valid_until).unwrap().as_secs() > 0 {
-        valid_domains.inner().remove(validation_id.as_str());
-        return Err(String::from("410"));
+    if now.duration_since(*valid_until).as_secs() > 0 {
```
fdb-hiroshima: You should remove the feature, as it's no longer used

src/routes/blogs.rs
```diff
@@ -100,25 +101,29 @@ pub fn new(rockets: PlumeRocket, _user: User) -> Ructe {
...
+        validation_map.remove(validation_id.as_str());
+        return Status::new(410, "validation expired");
```
fdb-hiroshima: Same as above

trinity-1686a (commented):
:eyes:

src/routes/blogs.rs
```diff
@@ -179,8 +179,24 @@ fn valid_slug(title: &str) -> Result<(), ValidationError> {
...
+fn valid_domain(domain: &str, valid_domains: State<Mutex<HashMap<String, Instant>>>) -> bool {
+    let mutex = valid_domains.inner().lock();
+    let mut validation_map = mutex.unwrap();
+
+    validation_map.insert(
+        utils::random_hex(),
+        Instant::now().checked_add(Duration::new(60, 0)).unwrap(),
+    );
+
+    true
```
fdb-hiroshima: I don't get returning a constant value, I don't see any early return?

src/routes/blogs.rs
igalic:
```
error[E0308]: mismatched types
   --> src/routes/blogs.rs:113:28
    |
113 |         return Status::new(Status::NotFound, "validation id not found");
    |                            ^^^^^^^^^^^^^^^^ expected u16, found struct `rocket::http::Status`
    |
    = note: expected type `u16`
               found type `rocket::http::Status`

error[E0308]: mismatched types
   --> src/routes/blogs.rs:123:28
    |
123 |         return Status::new(Status::Gone, "validation expired");
    |                            ^^^^^^^^^^^^ expected u16, found struct `rocket::http::Status`
    |
    = note: expected type `u16`
               found type `rocket::http::Status`
```

fdb-hiroshima: Have you tried simply
```suggestion
        return Status::Gone;
```
?

igalic: No, cuz I wanted to give a specific response.

fdb-hiroshima:
```suggestion
        return Custom(Status::Gone, "validation expired");
```
imported from rocket::response::status::Custom
Your version is sending a standard error code with a non standard text, some implementations don't handle this well. This return the corresponding "error" text, and set the body of the result to the text you provide

src/routes/blogs.rs
```diff
@@ -351,6 +351,10 @@ pub fn edit(name: String, rockets: PlumeRocket) -> Result<Ructe, ErrorPage> {
             .clone()
             .expect("blogs::edit: User was None while it shouldn't");
         let medias = Media::for_user(conn, user.id).expect("Couldn't list media");
+        let custom_domain = match blog.custom_domain {
+            Some(ref c) => c.to_string(),
+            _ => String::from(""),
```
AnaGelez: It is a quite personal opinion, but I think `blog.custom_domain.map(ToString::to_string).unwrap_or_default()` is more explicit (but, yes, it is only my opinion, maybe it is not).

igalic: another question might be: why were you able to use `Option<String>` for `theme` in `EditForm` but i have to use `String` here?

PR #598, opened 2019-05-25
refactor code to use Shrinkwraprs and diesel-derive-newtype

### LINKED ISSUE #597: Consider using Shrinkwrap for wrapper types
we have a lot of wrapper types in places — such as `Id(u32)`, or `SafeString(String)` which require a lot of boilerplate code.

Code that would be perfect for macro use!

Even better, when someone already made such macros: [shrinkwrap](https://crates.io/crates/shrinkwraprs)

igalic:
Reduce a lot of our own boilerplate with these two crates.
This fixes #597.


plume-models/src/reshares.rs
```diff
@@ -143,7 +143,7 @@ impl FromId<PlumeRocket> for Reshare {
                 post_id: Post::from_id(
                     c,
                     &{
-                        let res: String = act.announce_props.object_link::<Id>()?.into();
+                        let res = act.announce_props.object_link::<Id>()?;
```
BaptisteGelez: I think you can even replace this block with just `&act.announce_props.object_link::<Id>()?`, it was only here to be able to convert the Id to a String (and same for other similar blocks).

Issue #604, opened 2019-05-30
Mobile UX issues


GitHubGeek:
Issue 1: Classic editor - The Content textarea is not scrollable
Issue 2: Both editors - Too much left & right margain, squashing the editors

- **Plume version:** 59023e96023405847b3a91658e0adffecceb993d
- **Operating system:** Android 9 Pie
- **Web Browser:** Fennec F-Droid (Firefox android)

GitHubGeek:
Just noticed another issue: The link to _Classic editor (any changes will be lost)_ is also squashed on mobile. See screenshot:

![link squashed rich text editor mobile](https://user-images.githubusercontent.com/265551/58676275-fd563b80-839a-11e9-82d1-d9f11f9e1b20.png)

PR #609, opened 2019-06-02
Update textarea overflow to scroll

marek-lach:
Partial fix for #604.

### DIFF
--- a/static/css/_forms.scss
+++ b/static/css/_forms.scss
@@ -114,6 +114,7 @@ input[type="submit"] {
 //  Writing page
 form.new-post {
   max-width: 60em;
+  overflow-y: scroll;
   .title {
    	margin: 0 auto;
    	padding: 0.75em 0;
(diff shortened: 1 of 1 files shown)


marek-lach:
> I didn't tested (for some reason it is not deployed on pr-609.joinplu.me 😕), but it seems reasonable.

It does... I mean testing it live would probably be a good idea still, but there's really no reason why it should cause any problems. Pretty basic stuff overall...

static/css/_forms.scss
```diff
@@ -114,6 +114,7 @@ input[type="submit"] {
 //  Writing page
 form.new-post {
   max-width: 60em;
+  overflow-y: scroll;
```
BaptisteGelez: This line should be removed, otherwise there is a weird and useless scrollbar on the right of the whole form:

image

marek-lach:
Right. Done.

Issue #600, opened 2019-05-26
Password reset requests are forgotten when Plume is restarted


BaptisteGelez:
They are only stored in the app's memory for the moment, but they probably should have their own database table.

- **Plume version:** 0.3.0

PR #610, opened 2019-06-02
Store password reset requests in database

rfwatson:
Had a go at closing #600.

New to lots of stuff here (Rust, Diesel, Rocket, Plume...) so all feedback appreciated. 

## TODO

* [x] SQLite migrations
* [x] resolve inline TODOs
* [ ] integration tests?
* [x] fix CI issues

### DIFF
--- a/plume-models/src/password_reset_requests.rs
+++ b/plume-models/src/password_reset_requests.rs
@@ -1,4 +1,4 @@
-use chrono::{Duration, NaiveDateTime, offset::Utc};
+use chrono::{offset::Utc, Duration, NaiveDateTime};
 use diesel::{ExpressionMethods, QueryDsl, RunQueryDsl};
 use schema::password_reset_requests;
 use {Connection, Error, Result};
@@ -24,8 +24,8 @@ const TOKEN_VALIDITY_HOURS: i64 = 2;
 impl PasswordResetRequest {
     pub fn insert(conn: &Connection, email: &str) -> Result<String> {
         // first, delete other password reset tokens associated with this email:
-        let existing_requests = password_reset_requests::table
-            .filter(password_reset_requests::email.eq(email));
+        let existing_requests =
+            password_reset_requests::table.filter(password_reset_requests::email.eq(email));
         diesel::delete(existing_requests).execute(conn)?;
 
         // now, generate a random token, set the expiry date,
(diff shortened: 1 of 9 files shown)


src/routes/session.rs
```diff
@@ -239,43 +227,27 @@ fn passwords_match(form: &NewPasswordForm) -> Result<(), ValidationError> {
...
-                    i18n!(
-                        rockets.intl.catalog,
-                        "Sorry, but the link expired. Try again"
-                    ),
-                ))
-            }
+            let request = PasswordResetRequest::find_and_delete_by_token(&*rockets.conn, &token)
+                .map_err(to_validation)?;
+
+            // TODO show a better error if the token has expired?
```
rfwatson: The logic to detect an expired token is private to the model now. Not sure how important having a specific error is. We could maybe return a custom Diesel error?

plume-models/src/password_reset_requests.rs
```diff
@@ -0,0 +1,134 @@
...
+        });
+    }
+
+    #[test]
+    fn test_find_password_reset_request_by_token_time() {
+        let conn = db();
+        conn.test_transaction::<_, (), _>(|| {
+            user_tests::fill_database(&conn);
+
+            // TODO: best way to test this?
```
rfwatson: It would be good to test that old password reset requests are [ignored](https://github.com/Plume-org/Plume/pull/610/files#diff-a248869f1983a92462c39c8b2ea8a66eR44). 

The simplest way I can think of would be to insert a record, and then update its `creation_date` with some raw SQL in the test. It feels pretty dirty though. Better ideas appreciated.

BaptisteGelez: ~~`thread::sleep(2 * 60 * 60 * 1000)`~~

More seriously, I think you can pass tuples to `diesel::insert_into(...).values()`, like:

```rust
diesel::insert_into(password_reset_requests::table)
    .values((
        password_reset_requests::email.eq("foo@bar.org"),
        password_reset_requests::token.eq("aaaaaaaaa"),
        password_reset_requests::creation_date.eq(now - 3.hours()),
    ))
```

You can probably use it instead of `PasswordResetRequest::insert` in this test.

src/routes/session.rs
BaptisteGelez: Maybe, instead of mapping the error to `to_validation`, you could have a closure that returns the specific error that was previously returned?

rfwatson: I ended up adding a new `Expired` error variant - it felt generic enough to be useful elsewhere in the app too. Happy to iterate on this more if it doesn't feel like a good approach though.

rfwatson:
Pretty happy with this now. Still a couple of CI failures but they look random to me

BaptisteGelez:
Yes the CI fails sometimes because it doesn't have enough memory to build…

Issue #613, opened 2019-06-05
Refactor: only use Result in routes where an Error can occur


igalic:
Currently, we have a lot of code in our routes that uses `Result` as generic enum instead of a specific `#[derive(Responder)`, even tho it doesn't deal with errors at all:

```
plume-common/src/utils.rs:pub fn requires_login<T: Into<Uri<'static>>>(message: &str, url: T) -> Flash<Redirect> {
src/routes/session.rs:) -> Result<Flash<Redirect>, Ructe> {
src/routes/session.rs:pub fn delete(mut cookies: Cookies, intl: I18n) -> Flash<Redirect> {
src/routes/session.rs:) -> Result<Flash<Redirect>, Ructe> {
src/routes/reshares.rs:pub fn create_auth(blog: String, slug: String, i18n: I18n) -> Flash<Redirect> {
src/routes/instance.rs:) -> Result<Flash<Redirect>, Ructe> {
src/routes/instance.rs:) -> Result<Flash<Redirect>, ErrorPage> {
src/routes/instance.rs:pub fn ban(_admin: Admin, id: i32, rockets: PlumeRocket) -> Result<Flash<Redirect>, ErrorPage> {
src/routes/user.rs:pub fn me(user: Option<User>) -> Result<Redirect, Flash<Redirect>> {
src/routes/user.rs:pub fn dashboard_auth(i18n: I18n) -> Flash<Redirect> {
src/routes/user.rs:) -> Result<Flash<Redirect>, ErrorPage> {
src/routes/user.rs:) -> Result<Result<Flash<Ructe>, Redirect>, ErrorPage> {
src/routes/user.rs:pub fn follow_auth(name: String, i18n: I18n) -> Flash<Redirect> {
src/routes/user.rs:pub fn edit_auth(name: String, i18n: I18n) -> Flash<Redirect> {
src/routes/user.rs:) -> Result<Flash<Redirect>, ErrorPage> {
src/routes/user.rs:) -> Result<Flash<Redirect>, ErrorPage> {
src/routes/user.rs:) -> Result<Flash<Redirect>, Ructe> {
src/routes/notifications.rs:pub fn notifications_auth(i18n: I18n, page: Option<Page>) -> Flash<Redirect> {
src/routes/posts.rs:pub fn new_auth(blog: String, i18n: I18n) -> Flash<Redirect> {
src/routes/posts.rs:) -> Result<Flash<Redirect>, Ructe> {
[...]

PR #614, opened 2019-06-08
add enum containing all successful route returns

igalic:
This enum derives `Responder`, so it can be used as route result.
We also implement `From`, so it can be converted

This pr addresses #613

### DIFF
--- a/src/routes/blogs.rs
+++ b/src/routes/blogs.rs
@@ -236,104 +236,107 @@ pub fn update(
     name: String,
     form: LenientForm<EditForm>,
     rockets: PlumeRocket,
-) -> Result<Flash<Redirect>, Ructe> {
+) -> RespondOrRedirect {
     let conn = &*rockets.conn;
     let intl = &rockets.intl.catalog;
     let mut blog = Blog::find_by_fqn(&rockets, &name).expect("blog::update: blog not found");
-    if rockets
+    if !rockets
         .user
         .clone()
         .and_then(|u| u.is_author_in(&*conn, &blog).ok())
         .unwrap_or(false)
     {
-        let user = rockets
-            .user
-            .clone()
-            .expect("blogs::edit: User was None while it shouldn't");
-        form.validate()
-            .and_then(|_| {
-                if let Some(icon) = form.icon {
-                    if !check_media(&*conn, icon, &user) {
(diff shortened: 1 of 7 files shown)


igalic:
I've marked this pull request now ready for review.
even though there's at least one function left that could be improved (in readability, or at least refractored to use this enum)
but i don't think we can make further strides without improving our ErrorPage

src/routes/blogs.rs
```diff
@@ -111,42 +108,43 @@ pub fn create(
...
+            is_owner: true,
+        },
+    )
+    .expect("blog::create: author error");
+
+    Flash::success(
+        Redirect::to(uri!(details: name = slug.clone(), page = _)),
+        &i18n!(intl, "Your blog was successfully created!"),
+    )
+    .into()
```
rfwatson: So if I read this correctly, the Rust compiler infers the type from the return type of the function, allowing `into()` to function as expected?

Even though I should probably expect that from Rust, that's still pretty neat.

src/routes/session.rs
```diff
@@ -76,48 +77,43 @@ pub fn create(
...
+        Flash::success(
             Redirect::to(uri),
             i18n!(&rockets.intl.catalog, "You are now connected."),
-        ))
+        )
+        .into()
     } else {
-        Err(render!(session::login(
-            &rockets.to_context(),
+        render!(session::login(
```
fdb-hiroshima: Isn't it a bit strange that when redirecting to invalid uri, it gives an error? It would probably make more sens that the chain on line 100 does the parsing to, and default to `parse("/").unwrap()`. Also it should probably be a [Origin](https://api.rocket.rs/v0.4/rocket/http/uri/struct.Origin.html) to block redirection to other domain
(this is also true for previous code, it's pretty old, I'm pretty sure we did not do code review last September)

igalic: do you have a suggestion for improvement?

fdb-hiroshima: I do but I don't know how to explain it with words, so I'll probably do it myself during the week end

igalic: well, I'll be happy to review it then, and help with words where necessary

i used to be quite good with words, now i'm procrastinating on writing words, by working on a blogging platform.

Issue #611, opened 2019-06-04
Arm32 compile fails on nix dependency


mcrosson:
The arm32 build fail with the linked error. It seems this is fixed in an updated version of nix. Would it be possible for the nix version to be bumped to get arm32 support back?

Build errors: https://paste.lollipopcloud.solutions/?1185aefe37177bb9#uOSxiTy5YkA/ApGmpVHj0BZQ47equRKpVt1FkRHCAJM=

Nix issue: https://github.com/nix-rust/nix/issues/1057

mcrosson:
Looks like the below updates fix the arm32 builds. Would it be possible for someone to update the dependencies?

I'm not sure how to do it correctly for a PR otherwise I would have submitted a PR.

```
root@ff6e36226ece:/app# cargo update --package atomicwrites
    Updating crates.io index
    Updating atomicwrites v0.2.2 -> v0.2.3
    Updating libc v0.2.55 -> v0.2.58
    Updating nix v0.11.0 -> v0.14.1

root@ff6e36226ece:/app# cargo update --package ctrlc
    Updating crates.io index
    Updating ctrlc v3.1.2 -> v3.1.3
    Removing nix v0.13.0

cargo update --package nix
    Updating crates.io index

```

fdb-hiroshima:
if it works with just those 3 commands runned, you can git add Cargo.lock, commit and push, it should be enough

mcrosson:
Looks like the builds are back to green with arm32, closing the issue.

Thank you @fdb-hiroshima for all the help!

PR #615, opened 2019-06-13
Update dependencies which depended on nix -- fixes arm32 builds

mcrosson:
This addresses #611


fdb-hiroshima:
(don't worry about the CI failing, sometime when updating dependencies it run out of memory and fail, that's the problem when compiling on 36 cores with 2Gio of ram)

PR #619, opened 2019-06-18
Make a distinction between moderators and admins

### LINKED ISSUE #359: Allow to grant user admin right
There is currently no way to add/remove privileges from an existing user

### LINKED ISSUE #582: Differentiate between admin and moderator

There is no way to give someone access to the instance moderation tools without allowing them to change the instance settings too. Also it can be useful for users to be able to know who their admin and moderators are, to contact the right person if they have an issue.

A distinction between admin and moderator. Considering our needs, I think having a boolean `moderator` field on the `User` model will be enough, we don't need a full role/permission system.

Admins should be able to choose who becomes a moderator.

**~~Describe alternatives you've considered~~**

**Additional context**

[From Loomio](https://framavox.org/d/LUb3JMeX/moderation-tools)

AnaGelez:
And rework the user list in the moderation interface, to be able to run the same action on many users, and to have a huge list of actions without loosing space.

Peek 18-06-2019 18-36

Fixes #582, fixes #359


src/routes/instance.rs
```diff
@@ -199,27 +208,100 @@ pub fn admin_users(
...
+            "moderator" => Ok(UserActions::Moderator),
+            "un-moderator" => Ok(UserActions::RevokeModerator),
+            "ban" => Ok(UserActions::Ban),
+            _ => Err(()),
+        }
+    }
+}
+
+#[post("/admin/users/edit", data = "<form>")]
+pub fn edit_users(_mod: Moderator, form: LenientForm<MultiAction<UserActions>>, rockets: PlumeRocket) -> Result<Flash<Redirect>, ErrorPage> {
```
fdb-hiroshima: if I'm reading this right, and admin can't actually edit users, and a moderator can grant themselves admin rights

BaptisteGelez: Indeed.

I will try to make permissions an enum as you suggested.

For the UI part, the disadvantage of your proposal is that you can't apply the same action to a lot of users at the same time (but I don't know if it is really needed?).

fdb-hiroshima: maybe for mass-ban, but probably not for mass-granting-privileges

plume-models/src/users.rs
```diff
@@ -51,14 +51,44 @@ use {ap_url, Connection, Error, PlumeRocket, Result};
 
 pub type CustomPerson = CustomObject<ApSignature, Person>;
 
+#[derive(Clone, Debug, DbEnum)]
+#[DieselType = "User_role"]
+pub enum Role {
+    Normal,
```
igalic: perhaps, "Author"?

fdb-hiroshima: I don't know, this also includes someone who made an account just to comment

BaptisteGelez:
For translations, I think I will remove them for the moment. Once we will switch to the 2018 edition, I think we will be able to use the latest version of `gettext-macro` which has options to make smaller diffs.

plume-models/src/users.rs
```diff
@@ -138,6 +141,14 @@ impl User {
...
     find_by!(users, find_by_ap_url, ap_url as &str);
 
+    pub fn is_moderator(&self) -> bool {
+        self.role == 0 || self.role == 1
+    }
+
+    pub fn is_admin(&self) -> bool {
+        self.role == 0
+    }
+
```
igalic: can't we use an enum here instead of these numbers? (or in SQLite, a check constraint https://stackoverflow.com/questions/5299267/how-to-create-enum-type-in-sqlite)

plume-cli/src/users.rs
```diff
@@ -94,7 +100,17 @@ fn new<'a>(args: &ArgMatches<'a>, conn: &Connection) {
...
+
     let admin = args.is_present("admin");
+    let moderator = args.is_present("moderator");
+    let role = if admin {
+        0
+    } else if moderator {
+        1
+    } else {
+        2
+    };
```
igalic: that means, on this end, we'd just have a (one) `from_str()`

plume-models/src/users.rs
BaptisteGelez: See this commit message https://github.com/Plume-org/Plume/pull/619/commits/e3ecbdb2f4a8aa2b857057b27cd75b44c379767b

I tried to use a crate to do that with Diesel, but it generates a schema with the enum type for Postgres and with `Text` for SQlite, which we would have to maintain two different `User` struct depending on the database, one with `role: UserRole,` and the other with `role: String,`, and I don't want to do that.

plume-cli/src/users.rs
BaptisteGelez: Sorry, but… I don't understand. :sweat_smile: Could you try to rephrase, please?

plume-models/src/users.rs
fdb-hiroshima: Having an enum to make the comparison and not having magic numbers everywhere would still be cool
```rust
enum Role {
    Admin = 0,
    Mod = 1,
    User = 2,
}
fn main() {
    println!("{}", 0 == Role::Admin as u32)
}
```

BaptisteGelez: I didn't know it was possible to associate numbers to enum variants like that.

plume-models/src/users.rs
```diff
@@ -51,6 +51,12 @@ use {ap_url, Connection, Error, PlumeRocket, Result};
 
 pub type CustomPerson = CustomObject<ApSignature, Person>;
 
+pub enum Role {
```
igalic: if we `#[derive(Shrinkwrap)]` would it auto-convert `as i32`?

fdb-hiroshima: It's not a struct Role(i32), so I don't think so. Maybe tinkering with one of Deref, Borrow or AsRef can make it less verbose (like requiring only a & maybe)

fdb-hiroshima: thinking about it, [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4b2a159c18dbea965ddc2dc504571b93) is a thing, and is exactly what you hopped to do with Shinkwrap

PR #620, opened 2019-06-18
Use futures and tokio to send activities

### LINKED ISSUE #252: Improve ActivityPub delivery
When sending activities to a lot of other inboxes, we currently have two problems:

- some of them timeout, but blocks the delivery for a long time
- if one request fails, the following ones are never sent

We should try to see if we can make each request asynchronously, and make handle errors properly to avoid to stop the other tasks.

BaptisteGelez:
Fixes #252

Publishing an article and having it federated to the whole fediverse only takes a few seconds now (instead of minutes before, on big Plume instances connected to many other ones).

### DIFF
--- a/plume-common/src/activity_pub/mod.rs
+++ b/plume-common/src/activity_pub/mod.rs
@@ -1,6 +1,6 @@
 use activitypub::{Activity, Link, Object};
 use array_tool::vec::Uniq;
-use reqwest::ClientBuilder;
+use reqwest::r#async::ClientBuilder;
 use rocket::{
     http::Status,
     request::{FromRequest, Request},
@@ -8,6 +8,7 @@ use rocket::{
     Outcome,
 };
 use serde_json;
+use tokio::prelude::*;
 
 use self::sign::Signable;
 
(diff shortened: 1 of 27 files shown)


plume-common/src/activity_pub/mod.rs
```diff
@@ -129,38 +130,38 @@ where
         .sign(sender)
         .expect("activity_pub::broadcast: signature error");
 
+    let mut rt = tokio::runtime::current_thread::Runtime::new().expect("new rt");
+    let client = ClientBuilder::new()
+        .connect_timeout(std::time::Duration::from_secs(5))
+        .build()
+        .expect("Can't build client");
     for inbox in boxes {
         // TODO: run it in Sidekiq or something like that
```
igalic: is this comment still relevant?

plume-common/src/activity_pub/mod.rs
```diff
@@ -129,38 +130,38 @@ where
         .sign(sender)
         .expect("activity_pub::broadcast: signature error");
 
+    let mut rt = tokio::runtime::current_thread::Runtime::new().expect("new rt");
```
igalic: if this fails, we panic with "new rt".??

plume-common/src/activity_pub/mod.rs
```diff
@@ -129,38 +130,38 @@ where
...
-                } else {
-                    println!("Error while reading response")
-                }
-            }
-            Err(e) => println!("Error while sending to inbox ({:?})", e),
-        }
+                r.into_body().concat2()
+            })
+            .map(|response| {
+                println!("Response: \"{:?}\"\n\n", response)
```
igalic: i think we should talk about our logging strategy sometime 😅

plume-common/src/activity_pub/mod.rs
BaptisteGelez: Not really, but it is for the background jobs in general, for the moment we just spawn them in a separate thread and that's it. I will remove it and create a more general issue.

plume-common/src/activity_pub/mod.rs
BaptisteGelez: Hum… yes, I could probably change that. To be honest I copied this line from reqwest examples without really thinking about it :sweat_smile:

plume-common/src/activity_pub/mod.rs
BaptisteGelez: Indeed, we should probably use `log` and `env_logger` or something like that. Should I create an issue?

plume-common/src/activity_pub/mod.rs
```diff
@@ -129,38 +130,36 @@ where
...
+                .headers(headers.clone())
+                .header(
+                    "Signature",
+                    request::signature(sender, &headers)
+                        .expect("activity_pub::broadcast: request signature error"),
+                )
+                .body(body)
+                .send()
+                .and_then(move |r| {
+                    println!("Successfully sent activity to inbox ({})", inbox);
```
fdb-hiroshima: I think this print should be done in the map just under, otherwise output could get mixed quickly, making both logs useless

plume-common/src/activity_pub/mod.rs
```diff
@@ -129,38 +130,36 @@ where
...
+                    r.into_body().concat2()
+                })
+                .map(|response| println!("Response: \"{:?}\"\n\n", response))
+                .map_err(|e| println!("Error while sending to inbox ({:?})", e)),
```
fdb-hiroshima: (what is that comma for?)

BaptisteGelez: (no idea, I will remove it)

BaptisteGelez: Actually, it is added by cargo-fmt, because this whole expression is the argument of a function.

fdb-hiroshima: okay...

Issue #403, opened 2019-01-01
night mode


fdb-hiroshima:
From Matrix : 
> One minor thing: no night mode.

Some users prefer a darker theme, it would be great to have a simple switch somewhere on most pages to switch between two color themes, the one we already have, and a darker "night mode" one. Before implementing that, we should take into account how it should interact with #354 

Plume version (and/or commit): 0.2.0

igalic:
there's also this CSS draft https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme

AnaGelez:
Can we consider this as fixed now that we have a dark theme by default too? The only issue is that it doesn't respect `prefers-color-scheme`, but I don't really know if we could easily have an "automatic" version of the default theme, that switches between ligth and dark based on that.

igalic:
Chrome, Firefox and Safari already [support it](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme#Browser_compatibility)
as do most mobile browsers.
I think we can just add it, since CSS falls back gracefully (ignores things it doesn't know) we should be safe.

AnaGelez:
Oh, I had an idea, we could just use `@import` at the end of the CSS file to import the other variant when needed.

> although, i have no idea how to set my preference in my Firefox.

Probably some flag in about:config for the moment? Or maybe it uses the preference from your DE?

PR #624, opened 2019-06-21
Theming

### LINKED ISSUE #354: Allow customised blogs at user-level
A page or area where you can customise the way blog posts look for your (the user) posts only would be appreciated. A simple GUI and also (or alternatively) a custom CSS text area. This could enhance blogs and would be a great feature to add. WriteFreely are doing it successfully.

AnaGelez:
- Custom CSS for blogs
- Custom themes for instance
- New dark theme
- UI for admin to upload new theme (or delete them)
- UI to choose your theme

image

Instance level themes are just CSS files, stored in `static/css` (see `Instance::list_themes`).

TODO:

- [x] Federate blog themes
- [x] Escape url()/import
- [x] Add a way to avoid conflicts between blog and instance themes
- [x] UI to disable blog CSS (the option is already in the database, just need to add a checkbox in the settings)
- [x] SQlite migrations
- [x] Documentation

Fixes #354, #403


migrations/postgres/2019-06-19-141114_themes/up.sql
```diff
@@ -0,0 +1,4 @@
+-- Your SQL goes here
+ALTER TABLE blogs ADD COLUMN custom_css TEXT;
+ALTER TABLE users ADD COLUMN preferred_theme VARCHAR;
```
igalic: this should be a cookie setting

I really hope this is a cookie setting.

BaptisteGelez: Why should it be a cookie setting? Storing it in the database as the advantage of having it synced across devices (and also it will never expire).

igalic: you'd have to sync it to the cookie, but that way it'd be easier & quicker to access.

BaptisteGelez: The whole user struct is already loaded for every page anyway, I'm not sure it is going to make a real difference :woman_shrugging:

fdb-hiroshima:
is it still ready for review considering the design flaw pointed regarding arbitrary css security?

BaptisteGelez:
No, I'm implementing another approach, I think I will be able to push it this evening.

BaptisteGelez:
If you want to review this PR, the new version is there. Admins should now approve remote blog themes, but they may contain anything. Once a theme is approved, it is also available for use on this instance. Admins can also upload themes if they have SSH acces to their instance: they just need to create a folder called "blog-SOMETHING" in `static/css`, and add the theme files inside.

I will update the docs too.

trwnh:
Does it really make sense to have admins approve remote blog themes? Using remote resources on the local site seems fragile and prone to many more errors in the future. I think it makes more sense to limit themes to site-wide and local-user. 

More generally, this is an issue that has to do with the way articles federate in Plume -- the HTML is delivered and cached to many different sites that may have varying CSS. Federating CSS as well is, I think, out-of-scope of ActivityPub (although there is nothing preventing this). In any case, it is much simpler to say that users should view the original URL if they wish to read the article with the original CSS.

fdb-hiroshima:
I'm not sure this is an actual concern but how does it play with caching? Where are the file stored and are they given a unique name (hash based?)?

BaptisteGelez:
The themes are stored in `static/css` so they should integrate with the current caching system I think?

fdb-hiroshima:
I think they are cached, but not evicted when modified, so client might think the file was not updated when it was. Adding a timestamp, a hash, a random string or anything to the name (`myfile-DEADBEEF1312.css`) would fix it

BaptisteGelez:
I think it is going to be quite difficult, because in the current implementation, the main theme file should be named `theme.css`, and it may load other files of the theme, so we would have to find a way to detect the main theme file even if it has a hash in its name, and to edit all `url()`/`@import` inside of it to make sure other files are correctly loaded too… Maybe we can just set a default cache duration of one week or something like that, with a HTTP header? Or is it a problem not to have the latest theme update for a few days?

fdb-hiroshima:
Even 7 days is quiet long and will definitely be reported as a bug at some point. I'm at work currently, maybe we can discuss of it later on riot?

BaptisteGelez:
OK, feel free to ping me when you want to talk about it :)

migrations/sqlite/2019-06-21-154916_themes/down.sql
```diff
@@ -0,0 +1,88 @@
...
+    ap_url,
+    private_key,
+    public_key,
+    shared_inbox_url,
+    followers_endpoint,
+    avatar_id,
+    last_fetched_date,
+    fqn,
+    summary_html
+FROM users;
```
fdb-hiroshima: sqlite migration mi amor

plume-models/src/instance.rs
```diff
@@ -186,6 +186,60 @@ impl Instance {
...
+                        f.path()
+                            .file_name()
+                            .and_then(std::ffi::OsStr::to_str)
+                            .map(std::borrow::ToOwned::to_owned)
+                    })
+                    // Ignore the one starting with "blog-": these are the blog themes
+                    .filter(|f| !f.starts_with("blog-"))
+                    .collect()
+            })
+            .unwrap_or_default()
```
fdb-hiroshima: in case of error, return empty `Vec`? I think returning a `Result<...>` would make more sens, to have a proper error in case of.. error

plume-models/src/instance.rs
```diff
@@ -186,6 +186,60 @@ impl Instance {
...
+                        f.path()
+                            .file_name()
+                            .and_then(std::ffi::OsStr::to_str)
+                            .map(std::borrow::ToOwned::to_owned)
+                    })
+                    // Only keep the one starting with "blog-": these are the blog themes
+                    .filter(|f| f.starts_with("blog-"))
+                    .collect()
+            })
+            .unwrap_or_default()
```
fdb-hiroshima: Same goes

plume-models/src/users.rs
```diff
@@ -202,30 +204,6 @@ impl User {
             .map_err(Error::from)
     }
 
-    pub fn update(
```
fdb-hiroshima: so this whole function was useless? Some other must be too then

AnaGelez: Yes, I would like to do a big code cleanup one day, removing all `pub`, progressively reintroducing them until it compiles, and remove everything that Rust says is unused.

PR #633, opened 2019-07-09
use patched version of bitpacking with 1x and 8x disabled

### LINKED ISSUE #632: Problem building bitpacking v0.6.0
When running `cargo install --no-default-features --features postgres
`

```
error: failed to compile `plume v0.3.0 (/home/plume/Plume)`, intermediate artifacts can be found at `/home/plume/Plume/target


Caused by:
  Could not compile `bitpacking`.



Caused by:
  process didn't exit successfully: `rustc --crate-name bitpacking /home/plume/.cargo/registry/src/github.com-1ecc6299db9ec823/bitpacking-0.6.0/src/lib.rs --color always --crate-type lib --emit=dep-info,link -C opt-level=3 -C metadata=2470242703a6f858 -C extra-filename=-2470242703a6f858 --out-dir /home/plume/Plume/target/release/deps -L dependency=/home/plume/Plume/target/release/deps --extern crunchy=/home/plume/Plume/target/release/deps/libcrunchy-7e3361a81a9bb65a.rlib --cap-lints allow` (signal: 9, SIGKILL: kill)

```

- **v0.3.0:** 
- **Ubuntu 18.04:**

fdb-hiroshima:
This replace Tantivy's dependency on bitpacking by a version with unused features disabled. It make compilation of bitpacking require ~55% less memory, but is kinda dirty (although perfectly functional). I don't know if we should merge this, it mainly depends how fast will Tantivy bump version number to allow usage of the feature-able bitpacking.

fix #632

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,9 +249,9 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 [[package]]
 name = "bitpacking"
 version = "0.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
+source = "git+https://github.com/fdb-hiroshima/bitpacking?branch=disable-non-4x#ebda7708f56023e75a566b49f2165dd22d52f0dc"
 dependencies = [
- "crunchy 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
+ "crunchy 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


fdb-hiroshima:
I don't feel like "patching dependency of dependency" can be called "not that dirty". But if it's all ok with everybody, this do work as intended

igalic:
I 'd rather this was patched in bitpacking with feature flags, and then in tantivity, we use those feature flags.

i.e.: we did this the right way. we're not in a hurry, and we know how to get where we're going

fdb-hiroshima:
feature gating different packers has been merged to tantivy:master. Next release is coming shortly, so this pr does not make much sense

PR #646, opened 2019-08-21
Replace the input! macro with an Input builder

AnaGelez:
Pros:

- less ambiguities
- more type safety
- less cloning (we don't clone forms for each input anymore)

Cons:
- more verbose (but at the same time it is easier to parse IMO)
- less natural syntax


fdb-hiroshima:
Maybe implementing https://docs.rs/ructe/0.6.2/ructe/templates/trait.ToHtml.html would make more sens than a stand-alone function? Also, clippy is complaining about conventions, to_* usually take by reference

AnaGelez:
I thought about it, but my function takes a `gettex::Catalog`, while `ToHtml::to_html` doesn't. I could have passed it as an argument to `Input::new` tho.

But I made the commit before reading your review, so it is `Input::html` now (I can still change it if you prefer).

igalic:
definitely, what @fdb-hiroshima said

fdb-hiroshima:
I did not notice the Catalog. It was just a suggestion, but as it's totally fine as is

fdb-hiroshima:
As `.html()` take the Catalog, and the provided label in Input::new is _always_ translated, can't it be translated transparently from `.html()` instead of explicitly each time? (I don't know if it would be better or worst, that's an actual question)

AnaGelez:
gettext-macros requires strings that are known at compile time, so that it can add them to the .pot file. With the latest version of the crate (which requires Rust 2018 for some reason, so we can't use it now), there is a new macro, called `t` that would allow us to do:

```rust
Input::new("email", t!("Your email address"))
    .input_type("email")
    .html(catalog);
```

Which would be a bit better (but we can't really do anything else than that I think, these strings should be marked as translatable with a macro at some point).

src/template_utils.rs
```diff
@@ -247,98 +344,27 @@ macro_rules! input {
...
-            } else {
-                String::new()
-            },
-            val = escape(&$form.$name),
-            props = $props
+            details = self
+                .details
+                .map(|d| format!("<small>{}</small>", d))
+                .unwrap_or_default(),
+            error = self.error.unwrap_or_default(),
```
fdb-hiroshima: I think error is no longer put in a `<p class="error" dir="auto">{}</p>`

templates/blogs/new.rs.html
```diff
@@ -9,7 +9,11 @@
 @:base(ctx, i18n!(ctx.1, "New Blog"), {}, {}, {
   <h1 dir="auto">@i18n!(ctx.1, "Create a blog")</h1>
   <form method="post" action="@uri!(blogs::create)">
-      @input!(ctx.1, title (text), "Title", form, errors, "required minlength=\"1\"")
+      @(Input::new("title", i18n!(ctx.1, "Title"))
+      		.default(&form.title)
+      		.error(&errors)
+      		.set_prop("minlength", 1)
```
fdb-hiroshima: previously had the `required`property

templates/instance/admin.rs.html
```diff
@@ -17,7 +17,11 @@ <h1>@i18n!(ctx.1, "Administration")</h1>
   ])
 
   <form method="post" action="@uri!(instance::update_settings)">
-    @input!(ctx.1, name (text), "Name", form, errors.clone(), "props")
```
fdb-hiroshima: what was `props`, why is it gone?

templates/blogs/new.rs.html
fdb-hiroshima: mb, it's added automagically when not optional

templates/instance/admin.rs.html
AnaGelez: I don't know to be honest, but it clearly was here by mistake (this argument was for additional HTML properties).

Issue #642, opened 2019-08-07
The front-end is in Arabic if the user's language doesn't exist


AnaGelez:
If your browser is configured so that `navigator.language` is not one the language Plume's front-end is translated in, the front-end (mostly the editor) will be in Arabic. It should probably be in English instead.

The issue lays here: https://github.com/Plume-org/Plume/blob/master/plume-front/src/main.rs#L52 (`catalogs[0]` is the one for Arabic).

- **Plume version:** 0.3.0
- **Operating system:** Any
- **Web Browser:** Any

PR #648, opened 2019-08-22
Fix issue #642, frontend not in english if the user language does not exist

requiem958:
Fix [#642](https://github.com/Plume-org/Plume/issues/642)

Iterate over catalogs to find the english index and then use it instead of 0.

### DIFF
--- a/plume-front/src/main.rs
+++ b/plume-front/src/main.rs
@@ -47,9 +47,11 @@ lazy_static! {
         let lang = js! { return navigator.language }.into_string().unwrap();
         let lang = lang.splitn(2, '-').next().unwrap_or("en");
 
-        let english_position = catalogs.iter()
-	    .position(|(language_code,_)| *language_code == "en").unwrap();
-	catalogs
+        let english_position = catalogs
+            .iter()
+            .position(|(language_code, _)| *language_code == "en")
+            .unwrap();
+        catalogs
             .iter()
             .find(|(l, _)| l == &lang)
             .unwrap_or(&catalogs[english_position])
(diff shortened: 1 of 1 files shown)


plume-front/src/main.rs
```diff
@@ -46,10 +46,15 @@ lazy_static! {
         let catalogs = include_i18n!();
         let lang = js! { return navigator.language }.into_string().unwrap();
         let lang = lang.splitn(2, '-').next().unwrap_or("en");
+
+        let english_position = catalogs
```
igalic: this should be either a (cached?) constant, or a (cached?) config value

plume-front/src/main.rs
```diff
@@ -46,10 +46,15 @@ lazy_static! {
...
+            .iter()
+            .position(|(language_code, _)| *language_code == "en")
+            .unwrap();
```
igalic: while i understand that this is a simple operation, i don't think we need to do it every time

igalic: actually, never mind, this is an operation that's only dealing with things known at compile time

the optimiser will probably inline this as a constant literal

plume-front/src/main.rs
requiem958: It should be a constant yes, but the rust doc of constant says : 

> Constants can be set only to a constant expression and not to the result of a function call or any other value that will be computed at runtime.
So i thought the keyword const was forbidden in this case.

fdb-hiroshima: All this is in a lazy_static! call, so the result will be evaluated only one time, the first time CATALOG is accessed. The only thing it's not is cached between two different page loading, but that's not necessary imo

PR #658, opened 2019-08-27
delete notification on user deletion

### LINKED ISSUE #651: Follow notifications are not deleted when account is deleted
This issue needs to be confirmed, but you could normally reproduce by:

- creating an account "A"
- creating an account "B"
- following "A" with "B"
- deleting "B"
- going to the notifications page of "A"
- and it should crash, because the notification is still here but the associated account have been deleted.

As I said, this issue needs to be confirmed, I noticed that on fediverse.blog (which uses PostgreSQL), but it may only happen in very specific cases.

- **Plume version:** a6c84daa1a242246ac01260eb9bc201dd6e47830
- **Operating system:** Any
- **Web Browser:** Any

fdb-hiroshima:
fix #651

before merging, I'd like to add a regression test


plume-models/src/notifications.rs
```diff
@@ -64,6 +64,14 @@ impl Notification {
             .map_err(Error::from)
     }
 
+    pub fn find_followed_by(conn: &Connection, user: &User) -> Result<Vec<Notification>> {
+        notifications::table
+            .filter(notifications::kind.eq(notification_kind::FOLLOW))
+            .filter(notifications::object_id.eq(user.id))
```
AnaGelez: `object_id` is the ID of the `Follow` object actually.

fdb-hiroshima:
this has been stall for a bit, and I don't have time to fix the CI (the problem is not related to the fix, but to the way integration tests are run). I'll revert latests commit so we can merge it as is

Issue #621, opened 2019-06-19
order media by date DESC 


xmgz:
## media page enhacements

as it stands it has a good preview of the media but no so useful to select **and use** an image

### kind request 

- order media by date DESC so last uploaded is on top (likely you need to use it)
- consider show pictures on list format
- consider use "infinite scroll" or similar feature
- some "copy code" link/button to fast copy media's markdown code

**if first** option is implemented then 2 and 3 are not so needed

**4** is just for saying :P  low priority feature

### thanks

my guess is that just ordering db query to **select** images "by date desc" would not brake anything and it makes a difference

PR #660, opened 2019-08-27
order media so that latest added are top

fdb-hiroshima:
partial fix for #621

### DIFF
--- a/plume-models/src/medias.rs
+++ b/plume-models/src/medias.rs
@@ -82,6 +82,7 @@ impl Media {
     ) -> Result<Vec<Media>> {
         medias::table
             .filter(medias::owner_id.eq(user.id))
+            .order(medias::id.desc())
             .offset(i64::from(min))
             .limit(i64::from(max - min))
             .load::<Media>(conn)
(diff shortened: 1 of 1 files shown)


plume-models/src/medias.rs
```diff
@@ -62,7 +62,14 @@ impl MediaCategory {
 impl Media {
     insert!(medias, NewMedia);
     get!(medias);
-    list_by!(medias, for_user, owner_id as i32);
+
+    pub fn for_user(conn: &Connection, owner: i32) -> Result<Vec<Media>> {
+        medias::table
+            .filter(medias::owner_id.eq(owner))
+            .order(medias::id.desc())
```
igalic: is this guaranteed to be such?
do we have a create date anywhere?

fdb-hiroshima: we don't have date, and the ids we use are monotonic counter, unless we overflow that 64b signed counter (not likely to happen)

igalic: i wonder if anyone expects that editing a media pushes it to the top again…

AnaGelez: We can't edit media information for the moment, so…

igalic: well then… *pushes [resolve] button*

PR #662, opened 2019-08-29
Amend margins on mobile displays

### LINKED ISSUE #655: Margins are too narrow on mobile
This is not a high priority issue, but compared with the old theme, the left/right margins of a blog post are too narrowly close to the edges of the screen on mobile, especially on the 21:9 aspect ratio screens etc...

You can see the old margins pre ##624:

IMG_0101

And after:

IMG_0100

I have not seen any actual text being cropped just yet, but it'd be a good idea to have some small `margin` distance from the display's edges.

- **Plume version:** latest commit
- **Related issues:** #650
- **Operating system:** iOS 12/iPhone X, but I've also tested it on Android 9
- **Web Browser:** iOS Safari, and also Firefox and Chrome for Android

marek-lach:
Fixes #655 and also most of #604.

**Edit:** Better addressed by #663. Closing this issue now.

### DIFF
--- a/assets/themes/default/_forms.scss
+++ b/assets/themes/default/_forms.scss
@@ -144,7 +144,7 @@ form.new-post {
 
   & > * {
     flex-grow: 1;
-    max-width: 40%;
+    max-width: 60%;
   }
 }
 
--- a/assets/themes/default/_forms.scss
+++ b/assets/themes/default/_forms.scss
@@ -144,7 +144,7 @@ form.new-post {
 
   & > * {
     flex-grow: 1;
-    max-width: 60%;
+    max-width: 40%;
   }
 }
 
--- a/assets/themes/default/_forms.scss
+++ b/assets/themes/default/_forms.scss
@@ -115,7 +115,7 @@ input[type="submit"] {
 form.new-post {
   max-width: 60em;
   .title {
-   	margin: 0 auto;
+   	margin: auto 0;
    	padding: 0.75em 0;
 
    	background: none;
--- a/assets/themes/default/_article.scss
+++ b/assets/themes/default/_article.scss
@@ -170,7 +170,7 @@ main .article-meta {
     p {
       font-size: 1.5em;
       display: inline-block;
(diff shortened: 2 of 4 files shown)


AnaGelez:
Looks like it is too much now…

image

image

AnaGelez:
Still the same for me… :confused: I can try to have a look tomorrow or on Saturday if you want.

marek-lach:
I'll see what I can do...

marek-lach:
@AnaGelez All tested and seems to work well on my end.

AnaGelez:
The article content is indeed better (maybe even a bit too much), but some elements that shouldn't have margins now have (article cover/heading, and all the info at the bottom from the license to the comment area). :confused:

marek-lach:
> The article content is indeed better (maybe even a bit too much), but some elements that shouldn't have margins now have (article cover/heading, and all the info at the bottom from the license to the comment area). 😕

All of these issues should be ok now.

marek-lach:
Not really sure why the article cover image isn't fully scaled to be honest...

marek-lach:
> I felt you were struggling with CSS here

You are right. It's a lot of guess work for me, I am not really natural at it :-)

assets/themes/default/_forms.scss
```diff
@@ -144,7 +144,7 @@ form.new-post {
 
   & > * {
     flex-grow: 1;
-    max-width: 40%;
+    max-width: 60%;
```
marek-lach: This bit here seems to make the _article meta_ less squashed than before on small screens, I think?

Before:
Screenshot 2019-09-01 at 19 23 09

After #662:

Screenshot 2019-09-01 at 19 23 44
 
May be worth replicating in #663 if you want...

AnaGelez: I'm not sure these lines apply on *article meta* section, as it doesn't have the `split` class too (it is used only for the license/tags part, not the rest). But I have something similar to that in my PR. Thanks anyway!

PR #666, opened 2019-09-07
Add Snapcraft metadata and install/maintenance hooks

RAOF:
This is somewhat of a proof of concept; it currently:
 * Lets you configure `BASE_URL`, addresses, and such through `snap set plume base-url="blog.cooperteam.net"`
 * Generates a `ROCKET_SECRET` at install time, and
 * Runs the DB migrations on upgrade

Everything works, modulo the media upload directory not being configurable, and the base directory being read-only…


RAOF:
To test this you need to build the snap by running `snapcraft` in a checkout; this should make `plume_0.3.0_amd64.snap` (or whatever, as appropriate for your architecture). This can be installed with `sudo snap install --dangerous plume_0.3.0_amd64.snap`.

Minimal setup is then something like:
```
sudo snap set plume base-url=cooperteam.net db.type=sqlite
```
(only sqlite is actually supported at the moment).
While the snap will run migrations on *updates*, you still need to manually do the initial migration, with something like:
`sudo plume.plm migrate --path=/var/snap/plume/current`

igalic:
why is the installation `--dangerous`?

RAOF:
Because it's a local file (and you haven't pre-acknowledged its nonexistent signature), so it hasn't had any of the normal automated review. Once uploaded to the snap store that `--dangerous` will go away.

set-environment
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+#enabled="$(snapcraft get enabled)"
+#if [ -z "${enabled}" || "${enabled}" != "true" ]
+#    echo "Plume not yet enabled"
+#    exit -1
+#fi
```
igalic: why is this commented out?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,63 @@
...
+      cargo web deploy -p plume-front --release
+      cargo install --force --no-default-features --features sqlite --path . --root ${SNAPCRAFT_PART_INSTALL}
+      cargo install --force --no-default-features --features sqlite --path plume-cli --root ${SNAPCRAFT_PART_INSTALL}
+      cp -a assets migrations static target translations ${SNAPCRAFT_PART_INSTALL}
+      cp set-environment ${SNAPCRAFT_PART_INSTALL}
+    stage-packages:
+      - openssl
+      - libsqlite3-0
+
+  diesel-cli:
```
igalic: what do we need / usr diesel-cli for?

set-environment
RAOF: Because I forgot to remove it, apparently!

snap/snapcraft.yaml
RAOF: The [compile plume](https://docs.joinplu.me/installation/with/source-code) instructions say to install diesel_cli. Is that not necessary?

RAOF: The [compile plume](https://docs.joinplu.me/installation/with/source-code) instructions say to install diesel_cli. Is that not necessary?

set-environment
RAOF: It would be good to fix this up so that the server doesn't try to start until setup the been completed. This *was* a manual attempt at doing that, but didn't quite work.

snap/snapcraft.yaml
igalic: @fdb-hiroshima?

AnaGelez: It is for the 0.3.0 tag, but not for master (we have `plm migration run` now).

set-environment
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
```
AnaGelez: Could this file go in the `snap` directory too, or not?

snap/snapcraft.yaml
```diff
@@ -0,0 +1,63 @@
+name: plume
+base: core18
+version: '0.3.0' # just for humans, typically '1.2+git' or '1.3.2'
+summary: Multi-user blogging platform, federated over ActivityPub
+description: |
+  There is normally some text here; it should probably not be
```
AnaGelez: Maybe you are already aware of that, but just to have reminder: this description should probably be updated.

set-environment
RAOF: I *think* it can go in `snap/local`. I'll give it a try.

set-environment
RAOF: Done! Plume now doesn't try to run until it's been enabled, and doesn't let you enable it until you've set the minimum configuration items.

set-environment
RAOF: It can. Done!

snap/snapcraft.yaml
RAOF: Removed

snap/snapcraft.yaml
RAOF: Updated. Please give the verbiage a once-over!

RAOF:
This is now updated (sorry about leaving it hanging here).

You now need to set all the configuration options (eg: `sudo snap set plume base-url=cooperteam.net db.type=sqlite`) and then enable Plume (`sudo snap set plume enabled=true`). This will run the initial migrations, and so will be all ready to be configured with `plume.plm ...`.

PR #668, opened 2019-09-19
snap: *Another* fix for building

RAOF:


RAOF:
This has built correctly once, and is simpler, so hopefully this is the last attempt?

PR #677, opened 2019-10-15
Don't split sentences for translations

### LINKED ISSUE #675: Don't split sentences for translations
@ryonakano reported on Crowdin that this sentence was not translatable to Japanese because of the different ordering of the words in the Japanese grammar: https://github.com/Plume-org/Plume/blob/4b205fa995a6935a2512a9725d84635e889db48e/templates/session/password_reset_request_expired.rs.html#L8

There are probably a lot of sentences like that, and I think we could just have one string containing the HTML markup, because Crowdin is probably smart enough to hide it/display it as "untranslatable".

ryonakano:
Fixes #675

Also I've included the trailing period to the translatable string.

### DIFF
--- a/templates/session/password_reset_request_expired.rs.html
+++ b/templates/session/password_reset_request_expired.rs.html
@@ -5,5 +5,5 @@
 
 @:base(ctx, i18n!(ctx.1, "Password reset"), {}, {}, {
   <h1>@i18n!(ctx.1, "This token has expired")</h1>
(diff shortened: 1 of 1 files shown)


AnaGelez:
The string should be updated on Crowdin too, now!

ryonakano:
Screenshot from 2019-10-17 07-23-33

I should have escaped the href quotation marks :bowing_woman:

PR #681, opened 2019-10-23
Paginate the outbox responses. Fixes #669

### LINKED ISSUE #669: Paginate outboxes

For users with a lot of posts, Mastodon sometimes throws 422 errors when trying to fetch their profile.

The outboxes should be paginated to avoid being too long.

**Describe alternatives you've considered**

None, this is the standard way of doing it with AP.

**Additional context**

Originally reported by xosem@fediverse.blog (their profile is affected by this bug).

epsilon-phase:
This seems to work on our end, so let us know if there's anything else that needs fixing.


fdb-hiroshima:
There seems to be a few issues with you patch:
 - it returns the oldest posts first, which is the opposite of what other software are doing (but I'm not sure how much of an issue that actually is)
 - it does not link to previous/next page, so only 12 posts may be accessible. I'm not very familiar with the Activitypub crate, but I guess there is something about using `OrderedCollectionPage` instead/in addition to `OrderedCollection`

epsilon-phase:
That's a good start for sure then.

So it does need to also request that the most recent posts are retrieved in an earlier page too right?

epsilon-phase:
Okay, so, we think we understand what we need to do, but we're not sure what the proper kind of `Link` object is for this.

epsilon-phase:
With any luck that does it

plume-models/src/users.rs
```diff
@@ -320,13 +320,43 @@ impl User {
...
-
     pub fn outbox(&self, conn: &Connection) -> Result<ActivityStream<OrderedCollection>> {
-        let acts = self.get_activities(conn)?;
-        let n_acts = acts.len();
         let mut coll = OrderedCollection::default();
+        let first = ap_url(&format!("{}?page=1", &self.outbox_url));
+        let last = ap_url(&format!(
+            "{}?page={}",
+            &self.outbox_url,
+            self.get_activities_count(&conn) / 12
```
igalic: 12 is hard coded. for now.?

src/routes/user.rs
```diff
@@ -553,7 +556,15 @@ pub fn outbox(name: String, rockets: PlumeRocket) -> Option<ActivityStream<Order
     let user = User::find_by_fqn(&rockets, &name).ok()?;
     user.outbox(&*rockets.conn).ok()
 }
-
+#[get("/@/<name>/outbox?<page>")]
```
igalic: does a blog also have such an outbox that could benefit from such a pagination?

plume-models/src/users.rs
epsilon-phase: Oh, we didn't know about the constant elsewhere... But it's in the routes code rather than the models code.

We'll move it over then.

src/routes/user.rs
epsilon-phase: True enough, we'll get that made then.

plume-models/src/users.rs
```diff
@@ -320,13 +320,43 @@ impl User {
...
+    pub fn outbox_page(
+        &self,
+        conn: &Connection,
+        (min, max): (i32, i32),
+    ) -> Result<ActivityStream<OrderedCollectionPage>> {
+        let acts = self.get_activities_page(conn, (min, max))?;
+        let mut coll = OrderedCollectionPage::default();
         coll.collection_props.items = serde_json::to_value(acts)?;
-        coll.collection_props.set_total_items_u64(n_acts as u64)?;
+        //coll.collection_props.set_total_items_u64(n_acts as u64)?;
```
fdb-hiroshima: why is it commented? I think this can just be deleted

plume-models/src/users.rs
```diff
@@ -320,13 +320,43 @@ impl User {
...
+                "{}?page={}",
+                &self.outbox_url,
+                min / ITEMS_PER_PAGE + 1
+            ))))?;
+        coll.collection_page_props
+            .set_prev_link(Id::new(ap_url(&format!(
+                "{}?page={}",
+                &self.outbox_url,
+                min / ITEMS_PER_PAGE - 1
+            ))))?;
```
fdb-hiroshima: Previous page should not be added if we are already on the first page, and a next page should probable not be added either if `get_activities_page` returned less entries than it could have.

Also I think there is an out-of-one on the page index, if we are on page n°1 (that is the first with current implementation of `outbox` above), min is going to be 0, so next page will be 0/12+1 = n°1 again, and previous would be n°-1

plume-models/src/blogs.rs
```diff
@@ -220,12 +225,46 @@ impl Blog {
         coll.collection_props.items = serde_json::to_value(self.get_activities(conn)?)?;
         coll.collection_props
             .set_total_items_u64(self.get_activities(conn)?.len() as u64)?;
+        coll.collection_props
+            .set_first_link(Id::new(ap_url(&format!("{}?page=0", &self.outbox_url))))?;
+        coll.collection_props
+            .set_last_link(Id::new(ap_url(&format!(
+                "{}?page={}",
+                &self.outbox_url,
+                self.get_activities(conn)?.len() as u64 / ITEMS_PER_PAGE as u64
```
fdb-hiroshima: I think this does not give the right result, divisions are flooring where in this case we need them ceiling. An easy way to achive that is to add ITEMS_PER_PAGE-1 before dividing
```suggestion
                (self.get_activities(conn)?.len() + ITEMS_PER_PAGE - 1) as u64 / ITEMS_PER_PAGE as u64
```

plume-models/src/users.rs
epsilon-phase: That makes sense. Should be fixed now

plume-models/src/blogs.rs
epsilon-phase: Alright, that makes sense too.

plume-models/src/users.rs
epsilon-phase: That makes sense.

fdb-hiroshima: the part about not including previous/next when unnecessary is fixed, but I don't think you addressed the other half about page index

plume-models/src/users.rs
```diff
@@ -320,13 +320,46 @@ impl User {
             .load::<User>(conn)
             .map_err(Error::from)
     }
-
     pub fn outbox(&self, conn: &Connection) -> Result<ActivityStream<OrderedCollection>> {
-        let acts = self.get_activities(conn)?;
-        let n_acts = acts.len();
         let mut coll = OrderedCollection::default();
+        let first = ap_url(&format!("{}?page=1", &self.outbox_url));
```
AnaGelez: It starts at 1 for users, but 0 for blogs. Either are fine, but it would be better if it was consistent.

src/routes/blogs.rs
```diff
@@ -347,7 +347,15 @@ pub fn outbox(name: String, rockets: PlumeRocket) -> Option<ActivityStream<Order
...
 }
-
+#[allow(unused_variables)]
+#[get("/~/<name>/outbox?<page>")]
+pub fn outbox_page(
+    name: String,
+    page: Page,
+    _rockets: PlumeRocket,
+) -> Option<ActivityStream<OrderedCollectionPage>> {
+    None
```
AnaGelez: Could you please call `Blog::outbox_page` here, even if it doesn't do anything yet?

epsilon-phase: Sure thing.

plume-models/src/users.rs
epsilon-phase: Looks like it should be one given the current `Page` code.

Let's make it that then.

epsilon-phase:
@AnaGelez You wouldn't have any recommendations on fixing the retrieval of older messages would you?

AnaGelez:
The method that needs changing is this one: https://github.com/Plume-org/Plume/blob/master/plume-models/src/users.rs#L333

It should be modified so that it fetches all the pages, and puts all the activities in the returned `Vec`.

AnaGelez:
There is an issue with the `last`/`first`/`prev`/`next` links: the `https://` part is doubled (for instance I have `https://https://plume.one/@/admin_one/outbox?page=1`).

I think it is because `User::outbox_url` already includes `https://`, and thus the call to `ap_url` is superfluous, you can use `format!` directly.

epsilon-phase:
That's a silly mistake to have made. Should be fixed now.

AnaGelez:
OK, it now fetches the pages correctly, but it loops infinitely, because on the last page `last` point the current page (which, indeed, is the last).

PR #683, opened 2019-10-24
Add environmental variable to control path of media

epsilon-phase:
This was a request made on the matrix chat that will make it possible to release a snap version, that environmental variable being `MEDIA_UPLOAD_DIRECTORY`

### DIFF
--- a/plume-models/src/medias.rs
+++ b/plume-models/src/medias.rs
@@ -148,10 +148,12 @@ impl Media {
         if self.is_remote {
             Ok(self.remote_url.clone().unwrap_or_default())
         } else {
+            let p = Path::new(&self.file_path);
+            let filename: String = p.file_name().unwrap().to_str().unwrap().to_owned();
             Ok(ap_url(&format!(
-                "{}/{}",
+                "{}/static/media/{}",
                 Instance::get_local()?.public_domain,
-                self.file_path
+                &filename
             )))
         }
     }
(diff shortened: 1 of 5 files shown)


src/routes/medias.rs
```diff
@@ -72,7 +72,12 @@ pub fn upload(
                     .map(|ext| format!(".{}", ext))
             })
             .unwrap_or_default();
-        let dest = format!("static/media/{}{}", GUID::rand().to_string(), ext);
+        let dest = format!(
+            "{}/{}{}",
+            CONFIG.media_directory,
+            GUID::rand().to_string(),
```
igalic: is this supposed to be deterministic?

epsilon-phase: Is it deterministic? It seems to come up with new GUIDs each piece we upload.

igalic: aah, never mind, i was thinking of something else
(the deterministially random id we use per release for *our* assets)

plume-models/src/medias.rs
```diff
@@ -202,10 +204,9 @@ impl Media {
             .next()
             .map(ToOwned::to_owned)
             .unwrap_or_else(|| String::from("png"));
-        let path =
-            Path::new("static")
-                .join("media")
-                .join(format!("{}.{}", GUID::rand().to_string(), ext));
+        let path = Path::new(&super::CONFIG.media_directory)
+            .join("media")
```
fdb-hiroshima: the `media` part is already included in the default value of media_directory (`static/media`), and is, I think, not required for a custom media directory

fdb-hiroshima: the `media` part is already included in the default value of media_directory (`static/media`). I don't know if it should be removed from here or from there (it depends on how custom directory should appear), but I think there is one too much

epsilon-phase: Oh, thanks for the catch.

RAOF:
I've built a snap from this locally, setting the new `MEDIA_UPLOAD_DIRECTORY` appropriately, and it works!

PR #688, opened 2019-10-31
Add autosaving to the editor

### LINKED ISSUE #686: warn when leaving edit page with unsaved change
see [this toot](https://makerdon.org/@Sylphox/103041707542858913)

either an auto-saving feature, but it would probably require some form of versioning to not publish half a modification when editing an already published post, or simply a warning popup about losing changes when leaving the page (can be done easily in js, see [this anwser on stack overflow](https://stackoverflow.com/questions/7317273/warn-user-before-leaving-web-page-with-unsaved-changes))

~~**Describe alternatives you've considered**~~

~~**Additional context**~~

epsilon-phase:
This should fix #686 Confirmation prior to navigating away is easy enough to add, but it'd be good to know if it's wanted first since this seems like it might serve the same purpose

### DIFF
--- a/plume-front/src/editor.rs
+++ b/plume-front/src/editor.rs
@@ -67,23 +67,23 @@ impl From<stdweb::private::ConversionError> for EditorError {
 #[derive(Serialize, Deserialize)]
 struct AutosaveInformation {
     contents: String,
+    cover: String,
     last_saved: f64,
-    license:String,
-    subtitle:String,
-    tags:String,
+    license: String,
+    subtitle: String,
+    tags: String,
     title: String,
-
 }
 js_serializable!(AutosaveInformation);
-fn is_basic_editor()->bool{
-    if let Some(basic_editor) = window().local_storage().get("basic-editor"){
+fn is_basic_editor() -> bool {
+    if let Some(basic_editor) = window().local_storage().get("basic-editor") {
         basic_editor == "true"
-    }else{
+    } else {
         false
     }
 }
 fn get_title() -> String {
-    if is_basic_editor(){
+    if is_basic_editor() {
         return InputElement::try_from(document().get_element_by_id("title").unwrap())
             .ok()
             .unwrap()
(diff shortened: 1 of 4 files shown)


plume-front/src/editor.rs
```diff
@@ -100,6 +216,11 @@ fn filter_paste(elt: &HtmlElement) {
 }
 
 pub fn init() -> Result<(), EditorError> {
+    if let Some(ed) = document().get_element_by_id("plume-fallback-editor") {
+        window().set_timeout(autosave, 15000);
```
fdb-hiroshima: I guess set_interval does not exist? :confused:

plume-front/src/editor.rs
```diff
@@ -62,7 +64,121 @@ impl From<stdweb::private::ConversionError> for EditorError {
         EditorError::TypeError
     }
 }
-
+#[derive(Serialize, Deserialize)]
+struct AutosaveInformation {
+    contents: String,
+    title: String,
+    last_saved: f64,
+}
```
fdb-hiroshima: while content is probably the most annoying to loose, if we do a local save, we should as well save everything (including subtitle, tags...)

plume-front/src/editor.rs
epsilon-phase: It doesn't look like it, oddly.

plume-front/src/editor.rs
epsilon-phase: True enough, easy enough to add too

plume-front/src/editor.rs
epsilon-phase: Okay, works fine in the `js!`

plume-front/src/editor.rs
```diff
@@ -62,7 +69,183 @@ impl From<stdweb::private::ConversionError> for EditorError {
...
+fn get_cover() -> String {
+    let cover: SelectElement =
+        SelectElement::try_from(document().get_element_by_id("cover").unwrap())
+            .ok()
+            .unwrap();
+    console!(log, "Got cover");
+    cover.raw_value()
+}
+fn autosave() {
+    let info: AutosaveInformation = AutosaveInformation {
```
AnaGelez: Is the type hint really needed?

plume-front/src/editor.rs
```diff
@@ -62,7 +69,183 @@ impl From<stdweb::private::ConversionError> for EditorError {
...
+        .ok()
+        .unwrap();
+    tags.raw_value()
+}
+fn get_license() -> String {
+    let license: InputElement =
+        InputElement::try_from(document().get_element_by_id("license").unwrap())
+            .ok()
+            .unwrap();
+    console!(log, "Got license");
```
AnaGelez: Maybe this line should be removed ?

plume-front/src/editor.rs
```diff
@@ -62,7 +69,183 @@ impl From<stdweb::private::ConversionError> for EditorError {
...
+    license.raw_value()
+}
+fn get_cover() -> String {
+    let cover: SelectElement =
+        SelectElement::try_from(document().get_element_by_id("cover").unwrap())
+            .ok()
+            .unwrap();
+    console!(log, "Got cover");
```
AnaGelez: This one too

plume-front/src/editor.rs
```diff
@@ -62,7 +69,183 @@ impl From<stdweb::private::ConversionError> for EditorError {
...
+    if let Some(autosave_str) = window().local_storage().get(&get_autosave_id()) {
+        let autosave_info: AutosaveInformation = serde_json::from_str(&autosave_str).ok().unwrap();
+        let message = format!(
+            "{} {}?",
+            i18n!(
+                CATALOG,
+                "Do you want to load the local autosave last edited at"
+            ),
+            Date::from_time(autosave_info.last_saved).to_date_string()
+        );
```
AnaGelez: You should use `i18n!` for formatting, to allow the date to be placed somewhere else in the sentence if translators need to do so.

```rust
i18n!(
    CATALOG,
    "Do you want to load the local autosave last edited at {}";
    Date::from_time(autosave_info.last_saved).to_date_string()
)
```

plume-front/src/editor.rs
```diff
@@ -62,7 +69,183 @@ impl From<stdweb::private::ConversionError> for EditorError {
...
+        }
+    }
+}
+fn clear_autosave() {
+    window().local_storage().remove(&get_autosave_id());
+    console!(log, &format!("Saved to {}", &get_autosave_id()));
+}
+static mut AUTOSAVE_TIMEOUT: Option<TimeoutHandle> = None;
 
+unsafe fn autosave_debounce(interval: u32) {
```
AnaGelez: Why does it need to be `unsafe`?

plume-front/src/editor.rs
epsilon-phase: Oh, probably not

plume-front/src/editor.rs
epsilon-phase: Oh, it should probably just be unsafe on the lines that refer to the `static mut`

AnaGelez: No, I don't know much about unsafe, so maybe you are right and whole function should be. But also, I feel like there is a better way to do this, like going with lazy_static for instance.

epsilon-phase: That option looked off to us too tbh.. At least on its own
Maybe it'd be better to use a RefCell or something?

AnaGelez: Anything that will prevent having `unsafe` code is fine (not that it is really unsafe, we only have one thread when running in the browser).

plume-front/src/editor.rs
epsilon-phase: Didn't realize it was possible to do that with `i18n!`, that's nifty :)

PR #690, opened 2019-11-04
Fix the navigation menu not opening on touch

marek-lach:
Fixes the issue of the mobile menu not reacting to touch, in cases where mouse clicking is not being simulated.

### DIFF
--- a/plume-front/src/main.rs
+++ b/plume-front/src/main.rs
@@ -69,28 +69,10 @@ fn main() {
         .ok();
 }
 
-/// Toggle menu on mobile device
+/// Toggle menu on mobile devices
 ///
 /// It should normally be working fine even without this code
 /// But :focus-within is not yet supported by Webkit/Blink
-fn menu() {
-    if let Some(button) = document().get_element_by_id("menu") {
-        if let Some(menu) = document().get_element_by_id("content") {
-            button.add_event_listener(|_: ClickEvent| {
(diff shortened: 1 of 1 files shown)


AnaGelez:
As you can see [here](https://circleci.com/gh/Plume-org/Plume/9186?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), `add_event_listener` only takes one argument in Rust.

However, I think replacing `|_: ClickEvent|` with `|_: TouchEnd|` may work.

marek-lach:
The problem is that touchend is exclusive for iOS. Click event should still
probably stay for everything else...

AnaGelez:
It seems to work on all mobile browsers: https://developer.mozilla.org/en-US/docs/Web/API/Document/touchend_event and we don't really care about desktop here I think, right?

marek-lach:
Ok, I'll just modify it that way then...

cybergipsy2020:
Hi I noticed that the menu button on the top left button still won't respond on ios browsers(safari and firefox). My plume is 0.4.0 and ios is 13.5.1.

elegaanz:
I think Safari (and other browsers on iOS, which all use the same rendering engine, since Apple doesn't allow anything else in the AppStore) don't support WebAssembly. Maybe we should just switch back to pure JavaScript?

PR #695, opened 2019-11-15
Snap: store media in the proper directory

AnaGelez:
Didn't tested it, but it should be ok, if I understand how Snap works.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -45,4 +45,6 @@ then
     touch ${SNAP_COMMON}/initial-migrations-run
 fi
 
+mdkir "${SNAP_DATA}/media"
+
 exit 0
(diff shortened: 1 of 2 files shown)


RAOF:
Oh, damn. I didn't push this yet!

snap/hooks/configure
```diff
@@ -45,4 +45,6 @@ then
     touch ${SNAP_COMMON}/initial-migrations-run
 fi
 
+mdkir "${SNAP_DATA}/media"
```
RAOF: The `configure` hook gets run each time the snap is upgraded; the name might be a bit misleading. This would fail the first time the snap is upgraded, as `mkdir` will fail to create an existing directory.

It's also *slightly* misspelled :wink:

AnaGelez:
Closing in favor of #698

PR #698, opened 2019-11-16
Various snap fixes

RAOF:
This uses the new `MEDIA_UPLOAD_DIRECTORY` support to make media uploads work, cleans up some of the scripts, and sets the version based on git tags.


snap/snapcraft.yaml
```diff
@@ -1,6 +1,6 @@
 name: plume
 base: core18
-version: '0.3.0' # just for humans, typically '1.2+git' or '1.3.2'
+adopt-info: plume
```
igalic: does Snapchat build nightlies?

RAOF:
I *think* it's set up to build each commit. Certainly at least once a day.

PR #700, opened 2019-11-20
Fix #637 : Markdown footnotes

### LINKED ISSUE #637: Pandoc markdown and notes
The markdown syntax used in Plume is quite satisfactory, but there should be at least one small correction: adapt the syntax of the notes to the pandoc flavor. 

Usually, when writing markdown, most editors use the following syntax to create note calls and references:

```
Lorem ipsum[^note] partibus quidem

[^note]: This is a note.
```

However, with Plume, you have to use:

```
Lorem ipsum[^note] partibus quidem

[^note] This is a note.
```

The absence of the colon prevents you from writing in markdown in a classic editor, then importing the text into Plume without adjusting the notes. The opposite is also true.

papey:
Here is fix for issue #637 that : 

- Ensure footnotes classes are not filtered by html sanitizer
- Add some footnote style rules

An example of footnotes using this fix :

**Markdown :** 

```md
Here is something[^a]


Here is another thing[^b]

[^a]: something

[^b]: another thing
```

Footnotes rendeering

Note that there is some cpull-markdown related syntax : **a carriage return is needed between each note definition**

### DIFF
--- a/assets/themes/default/_article.scss
+++ b/assets/themes/default/_article.scss
@@ -481,3 +481,12 @@ input:checked ~ .cw-container > .cw-text {
     margin: auto $horizontal-margin;
   }
 }
+
+// Footnote related styles
+.footnote-definition {
+  p {
+    font-size: smaller;
+    // Make sur the definition is inline with the label-definition
+    display: inline;
+  }
+}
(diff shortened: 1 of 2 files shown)


papey:
Note that back reference are not supported (yet) by cpull-markdown : https://github.com/raphlinus/pulldown-cmark/issues/142

PR #704, opened 2019-12-10
remove use of rust in migration

### LINKED ISSUE #702: Migration 2019-04-28-201506_create_tantivy_index is failing to run
Updated to latest git from an old plume and the migration `2019-04-28-201506_create_tantivy_index` doesn't run (postgresql).

```
plume-dashie@fedistodon:~/Plume$ plm migration run
Configuration read from /home/plume-dashie/Plume/.env
Running migration 20190428201506
thread 'main' panicked at 'Failed to run migrations: Db(DatabaseError(__Unknown, "column users.role does not exist"))', src/libcore/result.rs:997:5
```

this is probably due to the migration using the latest Database Model, but with an old database, because `users.role` is created in migration `2019-06-18-152700_moderator_role` later.

fdb-hiroshima:
rewrite use_timelines_for_feed in pure sql
delete tantivy related migration. It will require reintroducing tantivy
initialisation in docs

fix #702


AnaGelez:
The CI is failing because of lettre, LGTM.

Issue #701, opened 2019-11-24
No theme at all on "Default theme"


Kekun:
I had the light theme, I modified my profile, the theme was "Default theme" and I didn't touch the combobox at all, I saved my changes, I had no theme.

When explicitly setting another theme, it works, when explicitely setting it back to "Default theme", I have no theme again.

![Capture d’écran de 2019-11-24 19-17-55](https://user-images.githubusercontent.com/2536339/69499169-918b0080-0eef-11ea-96eb-5892a33cef90.png)

- **Plume version:** 0.3.0
- **Operating system:** Fedora Silverblue 31.20191107.0
- **Web Browser:** GNOME Web 3.34.0-18-g574f7196c (current nightly)

marek-lach:
For context: https://bugs.launchpad.net/ubuntu/+source/epiphany-browser/+bug/173147

AnaGelez:
If this bug doesn't come from Plume but from GNOME Web, I'm closing this issue. If someone can reproduce with another browser/engine, I will reopen it.

Kekun:
The issue you refer to is:
- not on GNOME Web's bug tracker but on Ubuntu's,
- 12 years old,
- closed as invalid since more than 9 years.

And here is a screenshot of the same issue repeated with the very same steps in Firefox 70.0 as packaged in Fedora 31.

![Capture d’écran de 2019-12-10 09-18-47](https://user-images.githubusercontent.com/2536339/70508354-97f3bc00-1b2e-11ea-90d6-2048f1e34c53.png)

AnaGelez:
Ok, sorry you are right, I'm reopening.

Kekun:
Thanks! :)

marek-lach:
Very generic things that may, (or may not) help:
- Declaring the type="text/css" explicitly, linking to the theme's scss files
- Edit comments in stylesheets from // format to  /* ... */

verymilan:
It would be amazing to put a new release out, even if only for this, as it confuses the users quite a lot...

PR #707, opened 2019-12-16
Make the comment syntax consistent across all CSS

marek-lach:


marek-lach:
> I would like to know what is motivating this change, just more consistency? But I'm not opposed to it otherwise.

Well, that is the valid comment syntax for CSS, which ensures that no part of the styling will be accidentally skipped over by a browser upon rendering.

Issue #701 inspired this change, not because this change will solve that problem completely, but it may solve some part of it, and it doesn't hurt anything, so... why not...

AnaGelez:
The .scss files are normally compiled to minified .css files, and all the comments are removed in this process, so it shouldn't have any impact, but OK!

Issue #705, opened 2019-12-11
Add <link rel="canonical" href=… /> when an instance serves foreign content


edgarogh:
# Context

When I type my full name in Google, all the plume-related [results](https://www.google.com/search?q=edgar+onghena+inurl:plume) that appear are from other instances serving content from mine (articles, users). I have no problems with my content being accessible from other instances, it's the essence of federated networks after all but I had a little improvement to suggest regarding search results.

# Suggestion

Whenever an article/user page is served from a foreign instance, add a `<link rel="canonical" href=… />` pointing to the URL of the same content on the original instance. This tag makes complying search engines understand that the current page mirrors another one.

# Pros

 * Get rid of duplicate pages (or at least duplicate content) on search engines
 * Give more value to individual instances, instead of making them popular behind another one's search results
 * Prevent an instance from filling the first search results and progressively turning into the "entry point" to the Plume fediverse, which would create a kind of "monopoly", completely contrary to the goal of federated networks

_There might be cons, but I couldn't think of any..._

epsilon-phase:
This element should be placed in the `<head>` right?

epsilon-phase:
This should probably be closed now.

AnaGelez:
Indeed!

PR #708, opened 2019-12-16
Fix issue #705

epsilon-phase:
Add canonical links to user pages and articles.


epsilon-phase:
No notes about it being an issue on mozilla's page about link types

PR #709, opened 2019-12-16
Alpha 4

AnaGelez:


AnaGelez:
OK, so why aren't the translations here???

Issue #587, opened 2019-05-16
Banned emails list


AnaGelez:
**Is your feature request related to a problem? Please describe.**

Someone can quite easily create hundreds of fake email addresses and register as much accounts on Plume, to spam people. Someone who gets banned can also recreate an account with the same email.

**Describe the solution you'd like**

Being able to have a list of forbidden emails patterns would be useful. It should accept regex, wildcard patterns or something like that, so that if someone tries to generate emails on the fly following a pattern it can easily be stopped.

**~~Describe alternatives you've considered~~**

**Additional context**

[Discussion on Loomio](https://framavox.org/d/LUb3JMeX/moderation-tools)

BaptisteGelez:
Also, I'm thinking about it: we don't currently check if the email that we ask for during registration is valid, we probably should (but this would need its own issue).

epsilon-phase:
Would it suffice, for now, to provide some formatting checking?

It seems like an expedient thing to add while fiddling with the blocklist

igalic:
the easiest way to check for validity of an email address is to send a confirmation link to it

do we currently do that?

fdb-hiroshima:
The email is currently used _only_ for password reset. It's (laxly) checked to be valid on signup, but no mail is sent to it

igalic:
we should probably change that. some time in the future

PR #718, opened 2020-01-08
Email blocklisting

epsilon-phase:
This should address #587 and any concerns for the terminology hopefully.


templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,66 @@
...
+    @:base(ctx, i18n!(ctx.1, "Blocklisted Emails"), {}, {}, {
+    @tabs(&[
+        (&uri!(instance::admin).to_string(), i18n!(ctx.1, "Configuration"), false),
+        (&uri!(instance::admin_instances: page = _).to_string(), i18n!(ctx.1, "Instances"), false),
+        (&uri!(instance::admin_users: page = _).to_string(), i18n!(ctx.1, "Users"), false),
+        (&uri!(instance::admin_email_blocklist:page=_).to_string(),i18n!(ctx.1, "Email blocklist"), true)
+    ])
+    <form method="post" action="@uri!(instance::add_email_blocklist)">
+        @(Input::new("email_address",i18n!(ctx.1,"Email address"))
+        .details(i18n!(ctx.1,"The email address you wish to block. In order to block domains, you can use globbing syntax, for example '*@gmail.com' blocks all addresses from gmail.com"))
```
igalic: I'm a bit confused as to why there are no spaces after `, `s in function calls, starting here

epsilon-phase: Oh, that's because rustfmt can't touch the templates :)

Fixed.

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,66 @@
+@use templates::base;
+@use plume_models::blocklisted_emails::BlocklistedEmail;
+@use template_utils::*;
+@use routes::*;
+
+@(ctx:BaseContext, emails: Vec<BlocklistedEmail>, page:i32, n_pages:i32)
+    @:base(ctx, i18n!(ctx.1, "Blocklisted Emails"), {}, {}, {
+    @tabs(&[
```
AnaGelez: There should probably be a `<h1>` here, like on other administration pages.

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,66 @@
...
+        </label>
+        <input id="notify_user" type="checkbox" name="notify_user">
+        @(Input::new("notification_text", i18n!(ctx.1, "Blocklisting notification"))
+            .optional()
+        .details(i18n!(ctx.1, "The message to be shown when the user attempts to create an account with this email address")).html(ctx.1))
+        <input type="submit"  value='@i18n!(ctx.1, "Add blocklisted address")'>
+    </form>
+    <form method="post" action="@uri!(instance::edit_email_blocklist)">
+        <header>
+            <input type="submit"  value='@i18n!(ctx.1, "Delete blocklisted emails")'>
```
AnaGelez: This button should probably be red (`class="destructive"` IIRC), and explicitly say that it deletes all blocklisted emails if it is what it does.

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,66 @@
...
+        .details(i18n!(ctx.1, "The email address you wish to block. In order to block domains, you can use globbing syntax, for example '*@gmail.com' blocks all addresses from gmail.com"))
+        .set_prop("minlength", 1)
+        .html(ctx.1))
+        @(Input::new("note", i18n!(ctx.1, "Note")).optional().html(ctx.1))
+        <label for="notify_user">@i18n!(ctx.1, "Notify the user?")
+            <small>
+                @18n!(ctx.1, "Optional, shows a message to the user when they attempt to create an account with that address")
+            </small>
+        </label>
+        <input id="notify_user" type="checkbox" name="notify_user">
```
AnaGelez: This `<input>` should be before the label IMO, as it is usually done with checkboxes.

migrations/postgres/2020-01-05-232816_add_blocklist/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Your SQL goes here
+create table email_blocklist(id SERIAL primary key,
+                             email_address text ,
```
AnaGelez: Maybe make this column unique, to avoid duplicated blocks?

migrations/postgres/2020-01-05-232816_add_blocklist/up.sql
```diff
@@ -0,0 +1,6 @@
+-- Your SQL goes here
+create table email_blocklist(id SERIAL primary key,
```
AnaGelez: Little detail, so if you don't feel like doing it, don't change, but: keywords in SQL usually are in all caps.

plume-models/src/blocklisted_emails.rs
```diff
@@ -0,0 +1,145 @@
...
+            .count()
+            .get_result(conn)
+            .map_err(Error::from)
+    }
+    pub fn pattern_errors(pat: &str) -> Option<glob::PatternError> {
+        let c = Pattern::new(pat);
+        match c {
+            Err(e) => Some(e),
+            _ => None,
+        }
```
AnaGelez: You could use https://doc.rust-lang.org/std/result/enum.Result.html#method.err

plume-models/src/blocklisted_emails.rs
```diff
@@ -0,0 +1,145 @@
...
+    pub fn delete_entries(conn: &Connection, ids: Vec<i32>) -> Result<bool> {
+        use diesel::delete;
+        for i in ids {
+            let be: BlocklistedEmail = BlocklistedEmail::find_by_id(&conn, i)?;
+            delete(&be).execute(conn)?;
+        }
+        Ok(true)
+    }
+    pub fn find_for_domain(conn: &Connection, domain: &str) -> Result<Vec<BlocklistedEmail>> {
+        let effective = format!("%{}", domain);
```
AnaGelez: Maybe add a `@` so that in only matches one domain, and not all domains with the same ending (I doubt this case will ever be encountered actually, be just to be sure).

src/routes/instance.rs
```diff
@@ -174,6 +175,55 @@ pub fn admin_users(
...
+        let mut c: BlocklistEmailDeletion = BlocklistEmailDeletion { ids: Vec::new() };
+        for item in items {
+            let key = item.key.parse::<i32>();
+            if let Ok(i) = key {
+                c.ids.push(i);
+            }
+        }
+        Ok(c)
+    }
+}
```
AnaGelez: I'm almost sure I did something similar for user management, couldn't the code be shared between the two?

templates/instance/admin.rs.html
```diff
@@ -14,6 +14,7 @@ <h1>@i18n!(ctx.1, "Administration")</h1>
     (&uri!(instance::admin).to_string(), i18n!(ctx.1, "Configuration"), true),
     (&uri!(instance::admin_instances: page = _).to_string(), i18n!(ctx.1, "Instances"), false),
     (&uri!(instance::admin_users: page = _).to_string(), i18n!(ctx.1, "Users"), false),
+    (&uri!(instance::admin_email_blocklist: page=_).to_string(), i18n!(ctx.1, "Email blocklist"), false)
```
AnaGelez: This link is absent from the "Instances" page.

src/routes/instance.rs
```diff
@@ -174,6 +175,55 @@ pub fn admin_users(
...
+
+#[post("/admin/emails/new", data = "<form>")]
+pub fn add_email_blocklist(
+    _mod: Moderator,
+    form: LenientForm<NewBlocklistedEmail>,
+    rockets: PlumeRocket,
+) -> Result<Ructe, ErrorPage> {
+    BlocklistedEmail::insert(&*rockets.conn, form.0)?;
+    admin_email_blocklist(_mod, Some(Page(0)), rockets)
```
AnaGelez: It should probably redirect to `/admin/emails` instead.

src/routes/instance.rs
```diff
@@ -174,6 +175,55 @@ pub fn admin_users(
...
+    }
+}
+#[post("/admin/emails/edit", data = "<form>")]
+pub fn edit_email_blocklist(
+    _mod: Moderator,
+    form: Form<BlocklistEmailDeletion>,
+    rockets: PlumeRocket,
+) -> Result<Ructe, ErrorPage> {
+    BlocklistedEmail::delete_entries(&*rockets.conn, form.0.ids)?;
+    admin_email_blocklist(_mod, Some(Page(0)), rockets)
```
AnaGelez: It should probably redirect to `/admin/emails`.

src/routes/instance.rs
```diff
@@ -174,6 +175,55 @@ pub fn admin_users(
...
+            let key = item.key.parse::<i32>();
+            if let Ok(i) = key {
+                c.ids.push(i);
+            }
+        }
+        Ok(c)
+    }
+}
+#[post("/admin/emails/edit", data = "<form>")]
+pub fn edit_email_blocklist(
```
AnaGelez: Maybe call it `/admin/emails/delete` if it is what it does?

templates/instance/emailblocklist.rs.html
epsilon-phase: It deletes only the ones selected.

src/routes/instance.rs
epsilon-phase: We couldn't see how to do that cleanly, since that's significantly more complicated.

src/routes/instance.rs
epsilon-phase: That makes sense.

migrations/postgres/2020-01-05-232816_add_blocklist/up.sql
epsilon-phase: Might as well follow the conventions :)

plume-models/src/blocklisted_emails.rs
epsilon-phase: That's much nicer :)

src/routes/instance.rs
AnaGelez: OK, let's do it this way then, we can always refactor later if needed.

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,67 @@
...
+            </small>
+        </label>
+        @(Input::new("notification_text", i18n!(ctx.1, "Blocklisting notification"))
+            .optional()
+        .details(i18n!(ctx.1, "The message to be shown when the user attempts to create an account with this email address")).html(ctx.1))
+        <input type="submit"  value='@i18n!(ctx.1, "Add blocklisted address")'>
+    </form>
+    <form method="post" action="@uri!(instance::delete_email_blocklist)">
+        <header>
+            <input type="submit" class="destructive" value='@i18n!(ctx.1, "Delete selected emails")'>
```
AnaGelez: Actually, this button should probably be hidden if there is currently no blocked addresses, and a placeholder message should be shown (`<p class="center">` is made for that).

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,67 @@
...
+        (&uri!(instance::admin_users: page = _).to_string(), i18n!(ctx.1, "Users"), false),
+        (&uri!(instance::admin_email_blocklist:page=_).to_string(), i18n!(ctx.1, "Email blocklist"), true),
+    ])
+    <form method="post" action="@uri!(instance::add_email_blocklist)">
+        @(Input::new("email_address", i18n!(ctx.1, "Email address"))
+        .details(i18n!(ctx.1, "The email address you wish to block. In order to block domains, you can use globbing syntax, for example '*@example.com' blocks all addresses from example.com"))
+        .set_prop("minlength", 1)
+        .html(ctx.1))
+        @(Input::new("note", i18n!(ctx.1, "Note")).optional().html(ctx.1))
+        <input id="notify_user" type="checkbox" name="notify_user">
```
AnaGelez: Moving it inside of the `<label>` tag would be better, otherwise they are not on the same line, and it is hard to understand they are linked.

templates/instance/emailblocklist.rs.html
```diff
@@ -0,0 +1,67 @@
...
+        <label for="notify_user">@i18n!(ctx.1, "Notify the user?")
+            <small>
+                @18n!(ctx.1, "Optional, shows a message to the user when they attempt to create an account with that address")
```
AnaGelez: There is a `i` missing in `i18n` here, it doesn't show correctly (it shouldn't even compile, but for some reason, it does).

epsilon-phase: What an odd mystery

templates/instance/emailblocklist.rs.html
epsilon-phase: That makes sense

Issue #673, opened 2019-10-07
Missing date/id in atom feed


EliotBerriot:
Based on my research, entries included the Atom feed generated by Plume doesn't include any `atom:id` and `atom:published` elements.

While `published` is optional, `atom:id` looks mandatory in the [specification](https://tools.ietf.org/html/rfc4287#section-4.2.9) and its absence is likely to confuse feed readers.

In a personal projects of mine, I am aggregating multiple Plume feeds, and the absence of the `published` value means it's impossible to order the entries by date after aggregation.

Example from https://plume.social/@/Aster@amplifi.casa:

```xml

<feed xmlns="http://www.w3.org/2005/Atom">
    <title>Aster</title>
    <id>https://plume.social/@/Aster@amplifi.casa/atom.xml</id>
    <updated></updated>
    <entry>
        <title><![CDATA[Eu também ignorei essas coisas no começo]]></title>
        <id></id>
        <updated></updated>
        <author>
            <name>Aster</name>
            <uri>https://amplifi.casa/@/Aster/</uri>
        </author>
        <link href="https://amplifi.casa/~/Asterismos/eu-também-ignorei-essas-coisas-no-começo/" rel="alternate"/>
        <content type="html" src="https://amplifi.casa/~/Asterismos/eu-também-ignorei-essas-coisas-no-começo/">REMOVED</content>
    </entry>
</feed>
```

As you can see: 

- `id` is empty
- `updated` is empty
- `published` is missing

THere are probably a few more fields that are available  [in the spec](https://tools.ietf.org/html/rfc4287) you may want to include, like `summary`, `rights` or `category` (to indicate what blog the entry is coming from?)

papey:
Hi everyone, i want to work on this issue. Is [this function](https://github.com/Plume-org/Plume/blob/52d860d402e320b11f0da2700ef50e27b01f85d2/src/routes/user.rs#L605) the good place to start ? Thanks !

AnaGelez:
(sorry, for the delay) Yes it is, and there is a similar one in src/routes/blogs.rs IIRC.

Kekun:
This is very likely why I can't have my new fediverse.blog blog on Planet GNOME: https://gitlab.gnome.org/Infrastructure/planet-web/merge_requests/32.

papey:
Here is the PR attached to this issue : https://github.com/Plume-org/Plume/pull/720

papey:
The issue is partially fixed by the now merged PR #720.

[...]

PR #720, opened 2020-01-13
WIP: Fix RSS issues

papey:
Here is a PR relative to issue #673.

At this state, the PR adds `id` and `published` field for each blog entry inside the RSS feed.

I'm not sure but it's look like there is no method or field attached to `post` struct to get information about update metadata.

I can add this but I want to be sure that it's in the scope of this PR.

Thanks for you're reply.

### DIFF
--- a/src/routes/mod.rs
+++ b/src/routes/mod.rs
@@ -139,6 +139,7 @@ pub fn post_to_atom(post: Post, conn: &Connection) -> Entry {
                 })
                 .collect::<Vec<Person>>(),
         )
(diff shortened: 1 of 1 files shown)


epsilon-phase:
Our personal feeling is that you are correct in that adding an updated field is outside of the scope of the pull request.

We'll review this after dinner, thanks for contributing :)

papey:
Here is the issue related to update field for blog entries https://github.com/Plume-org/Plume/issues/722

PR #725, opened 2020-01-18
Update compiler

fdb-hiroshima:
update the compiler, and fix deprecation warnings & assimilated


fdb-hiroshima:
hum, it's broken when using ructe 0.6.4, but we get deprecation warnings (hard errors due to `-D all` when using clippy) when using an older version (0.6.2) :confused:

PR #727, opened 2020-01-21
Add warning for missing documentation

AnaGelez:
To force us to document our code.

I didn't added it to binaries (`src/` and `plume-cli/` because it didn't really made sense for them IMO), but I can.

Also, to keep this PR small, I propose that we slowly document each function as we modify them in other PRs. Does it seem reasonable to you?

### DIFF
--- a/plume-api/src/lib.rs
+++ b/plume-api/src/lib.rs
@@ -1,3 +1,5 @@
+#![warn(missing_docs)]
+
 #[macro_use]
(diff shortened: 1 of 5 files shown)


igalic:
can't we add that that somewhere to `clippy` or `check`?

AnaGelez:
Isn't there a better way to allow just these warnings in clippy?

PR #730, opened 2020-01-22
Go async

igalic:
update rocket and its dependencies to their `async` branch.
update our code to use rocket's async APIs


src/routes/mod.rs
```diff
@@ -171,34 +170,38 @@ pub mod well_known;
...
-                .ok()
-        } else {
-            Response::build()
-                .merge(self.0.respond_to(r)?)
-                .header(ETag(EntityTag::strong(etag)))
-                .ok()
-        }
+    fn respond_to(self, r: &'r Request<'_>) -> response::ResultFuture<'r> {
+        Box::pin(async move {
+            let contents = std::fs::read(self.0.path()).map_err(|_| Status::InternalServerError)?;
```
igalic: this std::fs::read could (should?) be async.

elegaanz: It could with something like async-std, yes. But maybe add a todo and leave it for a future PR?

PR #731, opened 2020-01-22
Snapcraft updates

RAOF:
Should now work again (:angry: )


snap/snapcraft.yaml
```diff
@@ -8,7 +8,7 @@ description: |
    * Media management: you can upload pictures to illustrate your articles, but also audio files if you host a podcast, and manage them all from Plume.
    * Federation: Plume is part of a network of interconnected websites called the Fediverse. Each of these websites (often called instances) have their own rules and thematics, but they can all communicate with each other.
    * Collaborative writing: invite other people to your blogs, and write articles together.
-grade: devel # must be 'stable' to release into candidate/stable channels
+grade: stable
```
fdb-hiroshima: what does that change implies?

RAOF: Technically, that we can promote the snap into the candidate & stable channels. Otherwise, it's just end-user messaging.

PR #732, opened 2020-01-24
snap: Fix build process for Tier 2 platforms

RAOF:
Unfortunately this only actually *fixes* the build for arm64;
it seems that we hit various bugs in Rust on ppc64 and armhf.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -31,21 +31,20 @@ parts:
       - pkg-config
       - libsqlite3-dev
       - gettext
-      - libclang-8-dev
-      - on arm64,armhf,ppc64el,s390x:
-        - lld-8
+      - libclang-9-dev
+      - lld-9
     override-build: |
       snapcraftctl set-version $(git describe --tags)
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -31,21 +31,20 @@ parts:
       - pkg-config
       - libsqlite3-dev
       - gettext
-      - libclang-8-dev
-      - on arm64,armhf,ppc64el,s390x:
-        - lld-8
+      - libclang-9-dev
+      - lld-9
```
igalic: maybe we should get rid of these dependencies again, so we can have broader platform support

RAOF: These actuality exist everywhere; I just dropped the (attempted) “*only* install lld on tier 2 platforms” logic.

With lld-9 the build failures on ppc64 and armhf are elsewhere.

elegaanz:
>  it seems that we hit various bugs in Rust on ppc64 and armhf.

should we wait for these bugs to be fixed, or can we merge now?

PR #741, opened 2020-02-24
Fixed #734 , passowrds are now hidden on the remote_interact pages

### LINKED ISSUE #734: Password is not hidden on remote interact pages
If you type your password on a "remote interact" page, anyone can read it, since the `type="password"` is not set on the `<input>` : https://github.com/Plume-org/Plume/blob/master/templates/remote_interact_base.rs.html#L24-L28

- **Plume version:** 0.4.0
- **Operating system:** Linux (but not relevant)
- **Web Browser:** Firefox (but not relevant I think)

ksteimel:
This is a quick fix to remove visible passwords on the remote_interact pages.

### DIFF
--- a/templates/remote_interact_base.rs.html
+++ b/templates/remote_interact_base.rs.html
@@ -25,6 +25,7 @@
 	    			.default(login_form.password)
 	    			.error(&login_errs)
 	    			.set_prop("minlength", 1)
+				.input_type("password")
 	    			.html(ctx.1))
 		        <input type="submit" value="@i18n!(ctx.1, "Log in")" />
 	    	</form>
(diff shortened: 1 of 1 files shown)


templates/remote_interact_base.rs.html
```diff
@@ -25,6 +25,7 @@ <h2>@i18n!(ctx.1, "I'm from this instance")</h2>
 	    			.default(login_form.password)
 	    			.error(&login_errs)
 	    			.set_prop("minlength", 1)
+				.input_type("password")
```
floreal, line 28: Just a little alignment issue
```suggestion
				    .input_type("password")
```

Issue #713, opened 2019-12-26
Indicator for boosted/liked posts in timelines


elegaanz:
**Is your feature request related to a problem? Please describe.**

It may be hard to understand why a post is in one of our timelines, especially in the case of someone boosting/linking someone else you don't follow, or when the same post appears multiple times because it got boosted.

**Describe the solution you'd like**

Adding a little "X boosted/liked" indicator to each concerned card

**Describe alternatives you've considered**

None?

**Additional context**

From #711

floreal:
I added them. Here is how post cards look now:
![image](https://user-images.githubusercontent.com/386442/79052495-2c823e00-7c37-11ea-84db-ac7be657c94c.png)

elegaanz:
The idea was more to have a "X shared this article", where X is the name of someone you follow. That way you would know why you are seeing this post in your timeline, even if you don't follow the original author. But your changes are still an improvement that would probably have been added at some point anyway, so they are welcome. :blush:

elegaanz:
Reopening because the PR didn't actually fixed this issue but another one.

floreal:
Oh didn't understood it that way. I'll make another PR then

PR #744, opened 2020-04-11
Fixed: #713: Disply likes and boosts on post cards

floreal:
This is a basic and simple way to display them. No interaction possible.

Alos, this patch is not optimised. as everytime a partial post_card is
called, `Post::count_likes()` and `Post::count_reshares()` are called
which means quite a few more quesies are sent to database unless diesel
uses some cache mechanisem.

A way to enhance this this would be to keep a count of likes and reshares
are kept in Post model / table.

### DIFF
--- a/assets/themes/default/_global.scss
+++ b/assets/themes/default/_global.scss
@@ -190,7 +190,28 @@ p.error {
 
  	background: $gray;
 
- 	text-overflow: ellipsis;
+  text-overflow: ellipsis;
+
+  footer.authors {
+    div {
+      float: left;
+      margin-right: 0.25em;
+    }
+
+    .likes { color: $red; }
+    .reshares { color: $primary; }
+
+    span.likes, span.resahres {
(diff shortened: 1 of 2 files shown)


floreal:
On dark theme:
image
On Light theme:
image

elegaanz:
Thanks for the PR.

I'm not 100% about the presentation, these counters looks too big and in the middle of the card to me, like if they were the most important information here. Maybe moving on the same line as the author/blog/date would help?

And for the like/share counter cache on Post, feel free to implement it if you want, but this PR can probably be merged as it is. I don't know if Diesel does caching, but Postgres and SQlite certainly do, so it should be fine.

floreal:
Ok, I'm not sure how I can adjust size of both feathers.svg is it possible to do this with scss styling ?

floreal:
Ok, how about that?

image

image

On larger cards:
image

PR #746, opened 2020-04-12
Fix #701 Preferable default theme

### LINKED ISSUE #701: No theme at all on "Default theme"

KitaitiMakoto:
This patch fixes #701 .

The issue occurs because form value of `email` is empty string `""` when user selects "Default theme" and it is set as preferred theme name. There's no theme named `""`, so any theme is not applied. By setting `NULL` to `users.preferred_theme` when "Default theme" is selected, instance theme is used as the user's preferred theme.

I used Diesel's `#[changeset_options(treat_none_as_null="true")]` attribute on `User` model. This may not be preferred but `save_changes` doesn't have option to set `NULL` sometimes and keep current value otherwise(See http://diesel.rs/guides/all-about-updates/ . It says "Diesel doesn't currently provide a way to explicitly assign a field to its default value"). This is the reason why I used that attribute, but it may have too wide inpact. Could you consider?

### DIFF
--- a/plume-models/src/users.rs
+++ b/plume-models/src/users.rs
@@ -54,6 +54,7 @@ pub enum Role {
 }
 
 #[derive(Queryable, Identifiable, Clone, Debug, AsChangeset)]
+#[changeset_options(treat_none_as_null="true")]
 pub struct User {
     pub id: i32,
     pub username: String,
--- a/plume-models/src/users.rs
+++ b/plume-models/src/users.rs
@@ -54,7 +54,7 @@ pub enum Role {
 }
 
 #[derive(Queryable, Identifiable, Clone, Debug, AsChangeset)]
-#[changeset_options(treat_none_as_null="true")]
+#[changeset_options(treat_none_as_null = "true")]
 pub struct User {
     pub id: i32,
     pub username: String,
--- a/src/routes/user.rs
+++ b/src/routes/user.rs
@@ -399,7 +399,7 @@ pub fn update(
         )
         .0,
     );
-    user.preferred_theme = form.theme.clone();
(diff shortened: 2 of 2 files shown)


KitaitiMakoto:
Thank you for merging!

verymilan:
Sadly a new user of mine still just ran into missing CSS on 0.5.0 :(

KitaitiMakoto:
@verymilan Can you re-save configuration again, please?

verymilan:
I am not sure what you mean, but as usual, i have made them selecting a theme and save, which was fixing their issue as always

KitaitiMakoto:
Sorry for my poor English. What I did mean is...

Can you have your users select "Default theme" and save it? It might be fix the issue.

Under the hood, their current theme is saved as empty string `""` but re-saving theme makes it `NULL`. And then default theme is applied.

verymilan:
Yes, i know, and this is how we did it - i was just expecting that this workaround wasn't needed anymore.

KitaitiMakoto:
Ah, sorry, I missed your point.

> i was just ~expecting~ assuming that this workaround wasn't needed anymore.

Yes, your realization is correct.

PR #747, opened 2020-04-14
Update translation files

KitaitiMakoto:
so that translators can work on Crowdin.

PR #748, opened 2020-04-14
Build app and pot files before upload pos to Crowdin

KitaitiMakoto:
so that `crowdin` command recognize updated `*.pot` files.

See #747 for details.


elegaanz:
I can't tell for sure if it will work, since this job only runs on master, but let's merge and see!

KitaitiMakoto:
Thank you! Now plume.pot on Crowdin has been updated!

PR #749, opened 2020-04-14
Move stylesheets for header from _global.sccss to _header.scss

KitaitiMakoto:
Hi,

I found menu icon is not present when I use small screen device(smart phone):

And wrote patch to fix it:

Could you consider to merge it?


KitaitiMakoto:
Thanks!

PR #751, opened 2020-04-17
Use LenientForm instead Form for CSRF protected pages

KitaitiMakoto:
Hi, 

I found that password reset request(`POST /password-reset`) and password reset(`POST /password-reset/<token>`) cause 422 Unprocessable Entity because Plume attempts to parse `csrf-token` in form to model fields. Using `LenientForm` instead of `Form` solved this problem.


KitaitiMakoto:
Thank you for meeting!

PR #752, opened 2020-04-18
Added: New translationss

floreal:
When building, new translations appear. This PR is here just to update translation files and update translations on crowdin.


elegaanz:
We usually do that only once before each release, but I can merge it if you want. Maybe our translation workflow is not ideal? And we should probably document it better…

floreal:
I don't know how Crowdin works, will merging it and sending translations through Circle CI delete al pending translations ?
If not, I think it's a good idea to send them before a release is made so that this could buy some time to translators to work before releases are made.
How other projects do that ?

elegaanz:
I don't really now how Crowdin works when you do that either…

KitaitiMakoto:
Hi, @floreal 
> will merging it and sending translations through Circle CI delete al pending translations ?

No, they won't delete any pending translations.

> If not, I think it's a good idea to send them before a release is made so that this could buy some time to translators to work before releases are made.

@elegaanz said [the reason why they merge translations just before new release](https://matrix.to/#/!EQCycwwoFqVhWztkfh:disroot.org/$1587278737846KfPpi:disroot.org?via=disroot.org&via=matrix.org&via=bau-ha.us):
> We usually pull translations just before publishing a release, to avoid polluting the git history with repetitive commits.

How do you think about this?

In an aside, you can download pending translations from Crowdin once you create an account.

1. Visit [project page](https://crowdin.com/project/plume)
2. Select your language
3. Open menu alongside file name: [Screenshot from Gyazo](https://gyazo.com/f9d20d45f3624af1539091e1db81ae53)
4. Tap "Download" link

After that, you can replacepo files with it and check translations on local app although it takes some time.

Issue #735, opened 2020-01-29
Atom feed does not validate


thorsten-panknin:
I noticed that the newsfeed for my blog throws errors which means it doesn't validate. I want to have my feed added to an aggregator which does not accept invaid feeds.

https://fediverse.blog/~/PerAsperaAdAstra/atom.xml

- **Plume version:** 0.4
- Operating system: Majaro Linux**
- **Web Browser:** Firefox 72.0.2

AnaGelez:
Judging by the errors that FeedValidator reports, this issue is mostly a duplicate of https://github.com/Plume-org/Plume/issues/673 and https://github.com/Plume-org/Plume/issues/703

The other errors are :+1: 
- `html` is not a valid MIME type
- A `atom:link` with `rel="self"` is apparently required
- `summary` is (sometimes only?) missing

PR #764, opened 2020-05-02
Fix Atom feed

KitaitiMakoto:
**This patches include breaking change.** I percent-encoded URI segments in `ap_url`s. This affects only future users, blogs and posts. Existing `ap_url`s will not be changed.

This fixes Atom feeds so that it conform to Atom spec. This should fix Atom-related issues #673 and #735.


elegaanz:
Could it be possible to encode URI only for the Atom feed, not in the whole app? The JSON-LD spec (on which ActivityPub is built) says that every URI actually is an IRI, meaning it should be able to contain any Unicode character (with very few exceptions), so I don't know how percent-encoded strings are treated…

KitaitiMakoto:
I've done! Thank you for your advice.

And... actually, Atom also uses IRI. https://tools.ietf.org/html/rfc4287#section-4.2.6 says:

> 4.2.6.  The "atom:id" Element
> (snip)
> Its content MUST be an IRI, as defined by [RFC3987].

This is the reason why I reverted all the code around percent-encoding. I don't know why [W3C feed validator][] claims IRIs are errors. But, how about deploying IRI version of Atom and convert them  to URI if someone reports issue about specified feed reader implementation?

[W3C feed validator]:https://validator.w3.org/feed/

elegaanz:
> But, how about deploying IRI version of Atom and convert them to URI if someone reports issue about specified feed reader implementation?

Sorry, but I'm not sure if I understand. Could you please try to rephrase ? (my English is so bad, help) :frowning_face:

KitaitiMakoto:
Thank you for reply. I'm also not good at English. My English should be hard to understand. Will rephrase it. What I mean was...

* My current patch allows Atom feeds to use IRI in `id`, `link` and so on.
* It is because Atom spec says their value is IRI.
* Some Atom reader implementations might not parse it correctly.
* I propose that we accept the possibility until someone reports the issue actually.
* How do you think about it? If it's okay, can you merge these patches?
* My point is these patches doesn't lose anything from now.

elegaanz:
Okay, it is clearer, thank you. Indeed, I think we can merge.

KitaitiMakoto:
Thank you!

PR #760, opened 2020-04-30
Fixes #745 Use plural form for 0 in French

KitaitiMakoto:
This fixes #745

PR #765, opened 2020-05-03
Keep atmarks and hashes in link

### LINKED ISSUE #745: i18n: when 0 likes or reshare displaied, translations says "One"
An Image worth more than a description: 
image
See [templates/posts/details.rs.html#L82](https://github.com/Plume-org/Plume/blob/master/templates/posts/details.rs.html#L82)

- **Plume version:** master

- **Operating system:** Linux

- **Web Browser:** Firefox

KitaitiMakoto:
I found hashes in link text like `[Fixes #745 Use plural form for 0 in French](https://github.com/Plume-org/Plume/pull/760)` are treated as hashtags and breaks the link. I think this is not preferable. How do you think?

### DIFF
--- a/plume-common/src/utils.rs
+++ b/plume-common/src/utils.rs
@@ -414,6 +414,7 @@ mod tests {
             ("not_a@mention", vec![]),
             ("`@helo`", vec![]),
             ("```\n@hello\n```", vec![]),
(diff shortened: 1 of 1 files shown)


KitaitiMakoto:
Thank you!

PR #767, opened 2020-05-04
Escape expressions in @Html

KitaitiMakoto:
I found "use your Fediverse account" link at https://plume.nixnet.xyz/~/BlogOwl/why-i-don't-have-a-personal-website is broken. This is because URI in HTML attributes is not escaped.

I know characters which need to be escaped is not allow in user name, so `escape(&uri!(user::details: name = &author.fqn).to_string()),` might be unnecessary. But escaping all expressions in HTML template is web application basis. For instance, in the future, special chars in user name might get allowed and at the time this link will be broken.


KitaitiMakoto:
Thank you!

PR #768, opened 2020-05-05
Make database connections configurable by environment variables

KitaitiMakoto:
so that we can tune performance.


plume-models/src/config.rs
```diff
@@ -193,6 +195,11 @@ lazy_static! {
             var("ROCKET_PORT").unwrap_or_else(|_| "7878".to_owned())
         )),
         db_name: DB_NAME,
+        db_max_size: var("DB_MAX_SIZE").map_or(None, |s| Some(
+            s.parse::<u32>()
+                .expect("Couldn't parse DB_MAX_SIZE into u32")
+        )),
+        db_min_idle: var("DB_MIN_IDLE").map_or(None, |s| s.parse::<u32>().ok()),
```
elegaanz, line 202: Is there a reason for panicking in one case and silently failing in the other?

KitaitiMakoto: Good catch! There's no reason why. I will rewrite this and push it.

KitaitiMakoto: Done.

KitaitiMakoto:
Thank you for the merge.

PR #771, opened 2020-05-17
Upgrade Tantivy to v0.12.0

KitaitiMakoto:
Hi,

I upgraded Tantivy to v0.12.0 because I want to introduce Japanese full-text search feature via [lindera-tantivy][], which requires Tantivy v0.12.0.

[lindera-tantivy]: https://github.com/lindera-morphology/lindera-tantivy


KitaitiMakoto:
> when rebasing, do i just skip this commit?
or do i need to do something special here?

To be honest, I don't know because I'm not familiar with Rust async. But I guess you need some work with this code at least calling `.await` and so on.

fdb-hiroshima:
I haven't looked at Tantivy for months but that looks in adequacy with changes that where on tantivy's issue tracker at that time :+1:

KitaitiMakoto:
Thank you!

PR #773, opened 2020-05-18
remove dependency on runtime-fmt

### LINKED ISSUE #769: remove dependency on runtime-fmt
(This came out of a discussion on IRC)

`runtime-fmt` uses perma-unstable rust APIs and is therefore susceptible to breakage (e.g. SpaceManiac/runtime-fmt#13, #674)

As far as I can tell, `rt_format!` is much more powerful than necessary and `src.replace("{uri}", replacement)` would work in both places it is used. `str::replace()` does not support formatting instructions such as `{uri:<100}`, so it's not a drop-in replacement. My guess is that only the literal `{uri}` actually needs to be replaced, though.

OddBloke:
Per the issue, "runtime-fmt uses perma-unstable rust APIs and is
therefore susceptible to breakage".

This replaces the two calls to rt_format! with .replace() and drops the
dependency.

Fixes #769

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2167,7 +2167,6 @@ dependencies = [
  "rpassword 4.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "rsass 0.9.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "ructe 0.9.2 (registry+https://github.com/rust-lang/crates.io-index)",
- "runtime-fmt 0.4.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


OddBloke:
I tested that this is the appropriate `replace()` invocation by writing the following temporary test:

```rust
mod tests {
    #[test]
    fn temp_test() {
        let template = format!(
            "https://{}/remote_interact?target={{uri}}",
            "some_public_domain"
        );
        assert_eq!(
            template.replace("{uri}", "substitute"),
            rt_format!(template, uri = "substitute").unwrap()
        );
    }
}
```

OddBloke:
Thanks for the merge!

PR #777, opened 2020-05-25
[WIP] [PoC] '&' -> '&mut'

jebrosen:
This is a (incomplete) proof of concept of a possible workaround for `Sync` issues that appear in `async` code.

Roughly, the issue is this:

* Route handler futures must be `Send`. This requirement comes from Rocket, and would be nontrivial and/or undesirable to change in Rocket.
* Handlers have an `&PlumeRocket` or an `&Connection` held across an `await` point
* Therefore, `&PlumeRocket` / `&Connection` must be `Send`
*  `&T: Send` iff `T: Sync`, so `PlumeRocket` / `Connection` must be `Sync`
* `PlumeRocket` contains a `Connection`, and `Connection` contains a diesel `PgConnection`, which is not `Sync`.

The approach demonstrated here is to change every `&PlumeRocket` or `&Connection` to an `&mut PlumeRocket` or `&mut Connection`. `&mut T` is `Send` if `T` is `Send`, so the problem is eliminated:

* Route handler futures must be `Send`.
* Handlers have an `&mut PlumeRocket` or an `&mut Connection` held across an `await` point
* Therefore, `&mut PlumeRocket` / `&mut Connection` must be `Send`
*  **`&mut T: Send` iff `T: Send`**, so `PlumeRocket` / `Connection` must be **`Send`**
* `PlumeRocket` contains a `Connection`, and `Connection` contains a diesel `PgConnection`, which **is `Send`.**

## Downsides
* In theory `&PlumeRocket` *could* allow more work to be done in parallel, at least in the future. It does not look like that is currently the case, since every call to the database blocks anyway.
* This change is pervasive - it reaches all the way to `FromId` and `Inbox`. I know relatively little about the overall structure of this code, so this could be incorrect or inconvenient in ways I don't know about!
* Many of the remaining errors are caused by or made worse by the `&` -> `&mut` change. A different solution that keeps `&` in more places would be easier to work with overall.
* This approach does not address the problem of making blocking database calls inside `async` fns, which can cause issues ranging from degraded performance to deadlocks.

## Alternatives
* Put a `Mutex` around the `Connection` somewhere. Uncontended mutexes (which this one should be) are not a *huge* performance concern, but `Mutex` may be at least as or more unwieldy than this solution throughout the code.
* Replace or wrap `Connection` with an API like `conn.run(|c| Post::load(&c)).await`, where `run` handles the synchronization. This has similar tradeoffs to a `Mutex`, is probably the most inconvenient option in terms of overall code changes, and is also a significant chunk of new code to write and debug. However, it has the advantage of being capable of fixing the blocking-in-`async`-fn problem.

### DIFF
--- a/plume-common/src/activity_pub/inbox.rs
+++ b/plume-common/src/activity_pub/inbox.rs
@@ -86,7 +86,7 @@ where
     /// - the context to be passed to each handler.
     /// - the activity
     /// - the reason it has not been handled yet
-    NotHandled(&'a C, serde_json::Value, InboxError<E>),
+    NotHandled(&'a mut C, serde_json::Value, InboxError<E>),
 
     /// A matching handler have been found but failed
     ///
@@ -139,16 +139,16 @@ where
     ///
     /// - `ctx`: the context to pass to each handler
     /// - `json`: the JSON representation of the incoming activity
-    pub fn handle(ctx: &'a C, json: serde_json::Value) -> Inbox<'a, C, E, R> {
+    pub fn handle(ctx: &'a mut C, json: serde_json::Value) -> Inbox<'a, C, E, R> {
         Inbox::NotHandled(ctx, json, InboxError::NoMatch)
     }
 
     /// Registers an handler on this Inbox.
     pub fn with<A, V, M>(self) -> Inbox<'a, C, E, R>
     where
-        A: AsActor<&'a C> + FromId<C, Error = E>,
+        A: AsActor<&'a mut C> + FromId<C, Error = E>,
         V: activitypub::Activity,
-        M: AsObject<A, V, &'a C, Error = E> + FromId<C, Error = E>,
+        M: AsObject<A, V, &'a mut C, Error = E> + FromId<C, Error = E>,
         M::Output: Into<R>,
     {
         if let Inbox::NotHandled(ctx, mut act, e) = self {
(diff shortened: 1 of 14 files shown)


fdb-hiroshima:
I think the best way to handle sql connections would be to have worker threads that are basically dedicated to that, and have a [mpsc channel](https://docs.rs/async-std/1.6.0/async_std/sync/fn.channel.html) through which requests can be send to them, alongside a [one shot channel](https://docs.rs/tokio/0.1.22/tokio/sync/oneshot/fn.channel.html) that allow to return a result. 
This is basically how actors work (like in Erlang and derivative, or the Actix lib for Rust), it would allow to keep `&`, would properly handle blocking operation out of async context, and _maybe_ allow to compile both Postgresql and Sqlite in the same binary (however this would also be a lot of work, not that much new code, but lots of moving things around)

jebrosen:
> I think the best way to handle sql connections would be to have worker threads that are basically dedicated to that

Yeah, I think that's more or less the direction I was going with "wrap `Connection` with an API like `conn.run(|c| Post::load(&c)).await`". I agree that it's a nicer overall solution, with the biggest drawback being:
> however this would also be a lot of work, not that much new code, but lots of moving things around

igalic:
> however this would also be a lot of work, not that much new code, but lots of moving things around

:woman_shrugging:

PR #781, opened 2020-06-15
Don't show boosts and likes for "all" and "local" in timelines

### LINKED ISSUE #711: Duplicated articles in timelines
When publishing an article, it gets duplicated in timelines after some time.

It looks like duplication happens every time someone likes or boosts the post (from the same instance, or from another one), but I'm not 100% sure.

- **Plume version:** 0.4.0
- **Operating system:** Linux, but not relevant IMO
- **Web Browser:** Firefox Nightly, but not relevant either

elegaanz:
The fix was quite easy actually :sweat_smile: I think it is just that it wasn't clear enough for us that `local`/`all` shouldn't match boosts and likes, but only new posts.

Fixes #711

### DIFF
--- a/plume-models/src/timeline/query.rs
+++ b/plume-models/src/timeline/query.rs
@@ -406,8 +406,8 @@ impl Bool {
                 }
             }
             Bool::HasCover => Ok(post.cover_id.is_some()),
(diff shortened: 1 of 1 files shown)


Chostakovitch:
Marvellous ! Thank you ! :D

PR #782, opened 2020-06-17
Add persian

elegaanz:


elegaanz:
OK, I broke the git history once again, why do I have these two other commits here???

Issue #783, opened 2020-06-17
Adding font for Persian language


DearRude:
Unfortunately, when you write in Persian language, the default system font will be displayed. Despite English, it's very unreadable with Persian.

I suggest using [Shabnam Font](https://github.com/rastikerdar/shabnam-font) for Persian interface and [Vazir Font](https://github.com/rastikerdar/vazir-font) for blogposts.
Vazir Font is a open-source font which is used in default in Mastodon, OpenStreetMap and Telegram for Persian/Arabic. It is also widely used in Persian blogging websites such as [Virgool](https://virgool.io/).

elegaanz:
I don't know exactly how we can load a different font depending on the language… Maybe we could add a `lang="xx"` to `<html>`, and then do something like :

```css
html[lang="fa"] {
    font-family: "Shabnam";
}
```

Anyway, thank you for your fonts suggestions!

DearRude:
I don't think if we can use a lang keyword for CSS fonts or not. However, based on what I saw on other websites, they use [Vazir WOL](https://github.com/rastikerdar/vazir-font/tree/master/dist/Farsi-Digits-Without-Latin) (without latin) in priority - so text triggers Vazir if Persian or iterates to the next font if not. 

I believe there is another way to specify a `unicode-range` when declaring Persian font so it only triggers certain characters.

theMasix:
#787

elegaanz:
I think we can close this issue, thanks to #787

PR #787, opened 2020-06-20
Add Vazir Font

theMasix:
I added the [Vazir Font](https://github.com/rastikerdar/vazir-font) based on a request from #783


assets/themes/default/dark.scss
```diff
@@ -4,6 +4,7 @@
 @import url('./fonts/Route159/Route159.css');
 @import url('./fonts/Lora/Lora.css');
 @import url('./fonts/Playfair_Display/PlayfairDisplay.css');
+@import url('./fonts/Vazir_WOL/Vazir_WOL.css');
```
igalic, line 7: what's the license are these files?

assets/themes/default/_variables.scss
```diff
@@ -24,14 +24,20 @@ $margin: 0 $horizontal-margin;
...
-$lora: "Lora", serif;
+$route159: "Vazir",
+"Route159",
+serif;
+$playfair: "Vazir",
+"Playfair Display",
+serif;
+$lora: "Vazir",
+"Lora",
+serif;
```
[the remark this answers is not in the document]
theMasix: VSCode Prettier extension did this.

assets/themes/default/dark.scss
theMasix: License added.

assets/themes/default/_variables.scss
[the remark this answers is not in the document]
igalic: since i'm not sure it's that much "prettier", or more readable for that matter, could we put this back into one line each?

after that it should be good to merge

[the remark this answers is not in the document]
theMasix: fixed.

assets/themes/default/_variables.scss
```diff
@@ -24,14 +24,14 @@ $margin: 0 $horizontal-margin;
 
 /* Fonts */
 
-$route159: "Route159", serif;
-$playfair: "Playfair Display", serif;
-$lora: "Lora", serif;
+$route159: "Vazir", "Route159", serif;
```
elegaanz, line 27: This font is for the interface. Maybe you could also import the Shabnam font as suggested in #783 instead of Vizir for this one? This can be done in another PR, but if you don't want to do it in this one, could you at least restore Route159? It looks like there are some margin issues with this font on some buttons (but not in other places strangely).

bildo

(it is quite subtle, but the text is not perfectly centered vertically)

igalic: could we do what was suggested and only enable this for certain Unicode ranges?

assets/themes/default/fonts/Vazir_WOL/Vazir_WOL.css
```diff
@@ -0,0 +1,65 @@
+@font-face {
+    font-family: Vazir;
+    src: url('Vazir-WOL.eot');
+    src: url('Vazir-WOL.eot?#iefix') format('embedded-opentype'),
+        url('Vazir-WOL.woff2') format('woff2'),
+        url('Vazir-WOL.woff') format('woff'),
+        url('Vazir-WOL.ttf') format('truetype');
+    font-weight: normal;
+    font-style: normal;
```
igalic, line 9: we could add `unicode-range` here
unfortunately, those being the same as Arabic, might break Arabic texts.
@DearRude, what do you think?

DearRude: Just like English and German, Persian and Arabic have common characters. We would not be able to cut down Arabic by `unicode-range`.
However, I suggest using Vazir and Shamim fonts for Arabic too. Since those fonts fully support Arabic and have been used in other projects in Arabic, I won't be a bad idea if we support both.

I can provide range of unicode in Persian and Arabic together if needed.

assets/themes/default/_variables.scss
DearRude: I believe using [non-Latin](https://github.com/rastikerdar/vazir-font/tree/master/dist/Farsi-Digits-Without-Latin) one would have the same effect. If we prioritize it, only Persian and Arabic chars will be effected.

assets/themes/default/fonts/Vazir_WOL/Vazir_WOL.css
igalic: the ranges are well documented, https://en.m.wikipedia.org/wiki/Arabic_script_in_Unicode but excessive
so i think we should use the non-Latin variants of the scripts.

theMasix: This ```Vazir-WOL``` is a non-latin package. But as @dearrude said, Persian and Arabic has some common characters.

igalic: WOL ➡️ Without Latin

so how come @elegaanz noticed those issues with the centering of the text? what's going on?

theMasix: As @elegaanz said, It's an strange issue and I don't know if ```Shabnam``` could fix it or not!

igalic: let's try!

theMasix: ```Shabnam``` Added. @igalic @elegaanz

PR #790, opened 2020-06-21
cargo release helper

igalic:
- add support for [cargo-release](https://github.com/sunng87/cargo-release)
- add & backfill Changelog!

----

how to use?

```shell
cargo release --dry-run --workspace minor
```

### DIFF
--- a/release.toml
+++ b/release.toml
@@ -1,0 +1,16 @@
+# we don't have a crate yet, so
+disable-publish = true
+# change when we all have gpg keys
+sign-commit = false
+dev-version-ext = 'dev'
+# update all crates in plume at once:
+consolidate-commits = true
+tag-prefix = ''
+
+pre-release-replacements = [
+  {file="CHANGELOG.md", search="Unreleased", replace="{{version}}"},
+  {file="CHANGELOG.md", search="\\.\\.\\.dev", replace="...{{tag_name}}", exactly=1},
+  {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}"},
+  {file="CHANGELOG.md", search="<!-- next-header -->", replace="<!-- next-header -->\n\n## [Unreleased] - ReleaseDate", exactly=1},
+  {file="CHANGELOG.md", search="<!-- next-url -->", replace="<!-- next-url -->\n[Unreleased]: https://github.com/assert-rs/predicates-rs/compare/{{tag_name}}...HEAD", exactly=1},
+]
--- a/release.toml
+++ b/release.toml
@@ -5,7 +5,6 @@ sign-commit = false
 dev-version-ext = 'dev'
 # update all crates in plume at once:
 consolidate-commits = true
-tag-prefix = ''
 
 pre-release-hook = ["crowdin", "pull"]
 
--- a/release.toml
+++ b/release.toml
@@ -7,10 +7,12 @@ dev-version-ext = 'dev'
 consolidate-commits = true
 tag-prefix = ''
 
+pre-release-hook = ["crowdin", "pull"]
+
 pre-release-replacements = [
-  {file="CHANGELOG.md", search="Unreleased", replace="{{version}}"},
-  {file="CHANGELOG.md", search="\\.\\.\\.dev", replace="...{{tag_name}}", exactly=1},
-  {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}"},
+  {file="CHANGELOG.md", search="Unreleased", replace="[{{version}}]"},
+  {file="CHANGELOG.md", search="\\.\\.\\.HEAD", replace="...{{tag_name}}", exactly=1},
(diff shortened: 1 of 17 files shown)


igalic:
uh… we'll need to add it as development dependency

release.toml
```diff
@@ -0,0 +1,16 @@
...
+# update all crates in plume at once:
+consolidate-commits = true
+tag-prefix = ''
+
+pre-release-replacements = [
+  {file="CHANGELOG.md", search="Unreleased", replace="{{version}}"},
+  {file="CHANGELOG.md", search="\\.\\.\\.dev", replace="...{{tag_name}}", exactly=1},
+  {file="CHANGELOG.md", search="ReleaseDate", replace="{{date}}"},
+  {file="CHANGELOG.md", search="<!-- next-header -->", replace="<!-- next-header -->\n\n## [Unreleased] - ReleaseDate", exactly=1},
+  {file="CHANGELOG.md", search="<!-- next-url -->", replace="<!-- next-url -->\n[Unreleased]: https://github.com/assert-rs/predicates-rs/compare/{{tag_name}}...HEAD", exactly=1},
```
elegaanz, line 15: Shouldn't the repo URL be changed here?

elegaanz:
Also, maybe we could unify the headers in the changelog to follow what https://keepachangelog.com/ recommends, to keep it coherent.

And for the development dependency, I don't think so. It should be documented [here](https://docs.joinplu.me/organization/release-process/) that people who want to publish a new release should use this tool tho.

release.toml
igalic: how?

elegaanz: To something like `https://github.com/Plume-org/Plume/compare/{{tag_name}}...HEAD`, so that the changelog links to our commit history.

elegaanz: Also, is there a `<!-- next-url -->` somewhere in the changelog you added? I can't find it.

igalic:
n. b. : i've back-filled the changelog by copy-pasting the text from our github releases page

for the future of the changelog, i absolutely agree

elegaanz:
I know you copied what is on GitHub, but maybe we could change this text? Would it be a problem to have texts that are a bit different on Github and in the changelog? I would prefer to have the whole changelog have the same format (but that's arguable).

elegaanz:
Also, could it be possible to add `pre-release-hook` to run `crowdin pull` (I don't remember the exact arguments, but basically : the command to pull the latest version of translations)? Or do you think it should be done manually before running `cargo release`?

igalic:
> I know you copied what is on GitHub, but maybe we could change this text? Would it be a problem to have texts that are a bit different on Github and in the changelog? I would prefer to have the whole changelog have the same format (but that's arguable).

the only problem was time and energy
y'all have access to write to this PR, so, please, by all means, help me!!

> Also, could it be possible to add `pre-release-hook` to run `crowdin pull` (I don't remember the exact arguments, but basically : the command to pull the latest version of translations)? Or do you think it should be done manually before running `cargo release`?

that's a very good idea!

how do i make it work?

```shell
◆ plume git:(fix/cargo-release-helper) ❯❯❯ crowdin pull
Project key is empty
✖ 1 plume git:(fix/cargo-release-helper) ❯❯❯ 
```

elegaanz:
I don't remember how the Crowdin CLI works, I'll check that and edit the PR. I'll also edit the changelog.

elegaanz:
So the translations should be pulled now, if you have a `CROWDIN_API_KEY` in your env. We should remember to document how to get one before making a release.

igalic:
> So the translations should be pulled now, if you have a `CROWDIN_API_KEY` in your env. We should remember to document how to get one before making a release.

let's create a doc issue right now, and link it here

elegaanz:
https://github.com/Plume-org/docs/issues/88

igalic:
what about docker?

elegaanz:
I think the release should be automatic, I'll check.

elegaanz:
It is not, I'll a hook.

elegaanz:
The thing is, the docker tag needs to be pushed after the git tag, and there is no `post-release-hook` in cargo release, apparently.

igalic:
or we could add docker tag to the CI build process, only on tag

igalic:
yunohost relies on docker, right?
and it's one of *the* platforms that we recommend, thru our installation guide

elegaanz:
It doesn't, it uses Bash scripts. I'm not sure if it is better tho… :sweat_smile:

release.toml
igalic: i didn't add it to our current changelog

igalic:
we should open a pr to [release a new version on Yunohost](https://github.com/YunoHost-Apps/plume_ynh) after the release?

KitaitiMakoto:
That's good!

Issue #785, opened 2020-06-19
Specify preferred language


DearRude:
Realized that Plume language interface is selected with browser language. This works pretty well but since Plume is federated, the probability for an instance to be regional is high; hence It's possible for an instance to be mostly used in one language.

It would be so useful if you could select interface language (or auto for normal way) of an instance in the `.env` file. 

I tried to implement it myself but since I couldn't understand how `lazy_static!` macro works, I was not able to contribute.

DearRude:
Any ideas?

igalic:
@DearRude why don't you open whatever you had as WIP PR, and we'll fix it up into usable shape together.!

DearRude:
Okay!

PR #792, opened 2020-06-23
Force interface language by config file

DearRude:
As explained in #785, trying to force the language interface which specified in `.env` file.

I tried the commented code below and re-compiled the front, but nothing happened. Since I didn't find any other piece which gets browser language, I assume my method of recompilation is flawed. (Only do `cargo web deploy -p plume-front --release`)

So I made this WIP PR as @igalic suggested. 😁

### DIFF
--- a/plume-front/src/main.rs
+++ b/plume-front/src/main.rs
@@ -46,6 +46,9 @@ lazy_static! {
         let lang = js! { return navigator.language }.into_string().unwrap();
         let lang = lang.splitn(2, '-').next().unwrap_or("en");
 
+        // Force a language (Add .env setting)
+        //let lang = "fa"; ???
+
         let english_position = catalogs
             .iter()
             .position(|(language_code, _)| *language_code == "en")


elegaanz:
The issue with the front-end is that it runs in the browser: it can't access the environment variables of the server.

I think the solution would be to:
- first, do this change on the server side, where you can access the environment variables from the configuration. This may require you to change Plume-org/rocket-i18n first
- then, once this settings works on the server, add a `lang` property to every `<html>` tag we render, containing the language code that we chose to render the page
- in the front-end, use this `lang` attribute to detect the language instead of relying on `navigator.language` as we currently do.

I hope this isn't too much. I can try to guide you more if need detailed steps, I know it can be hard to get started on a codebase you know nothing about. Feel free to ask me here or on Matrix if you have any questions.

trinity-1686a:
> I assume my method of recompilation is flawed. (Only do `cargo web deploy -p plume-front --release`)

The feature is missing, so it's not the issue at hand, but doing just that was actually not sufficient. You must also rebuild Plume, as it's build script (build.rs) is the one copying files from `target/` to where static files should be

DearRude:
I give up on this. Could not make it work...

Issue #786, opened 2020-06-19
add caching headers for posts


igalic:
until we've switched to using the async version of Rocket, we'll have serious performance issues.

Most installations of Plume are already behind a (caching) proxy server. we can use this to fix the current performance issues by sending caching headers for posts pages:

### How?

the posts page should send an `ETag` with each response.
The advantage here is that we can have updates, and comments, since this means that we send new `ETag` header.
to spare a lookup in the database, we could store data about whether the post's cache is invalidated within rocket (that would be a `304` response to a `HEAD` request)

n.b.: The (caching) proxy server admin could adjust the time themselves, we wouldn't be sending any ourselves.

### Alternatives

are currently a lot more effort:

- #730 
- #777 

and there's also: https://github.com/SergioBenitez/Rocket/pull/1343
which is similar to #777, but on rocket level

those aren't really alternatives, since this suggestion is operating at HTTP level, it's more orthogonal.

elegaanz:
We already have some kind of caching for static assets : https://github.com/Plume-org/Plume/blob/master/src/routes/mod.rs#L207-L270

But adding it to templates could improve performances too.

igalic:
does rocket derive `#[head()]` for these routes itself?

elegaanz:
[Yes](https://rocket.rs/v0.4/guide/requests/#head-requests)

igalic:
if we need a DB lookup to construct a body which might still be in a downstream cache, then we should define `HEAD` ourselves

elegaanz:
Maybe we can have a `Vec<String>` storing the slugs of the posts that need an update in Rocket's state, and only show an updated version when needed.

Another solution would be to render each post to a given HTML or AP file when the post or comments change, and serve this files statically, using the modification date to generate the cache headers.

trinity-1686a:
There is already caching for any templated page, with the following limitations : 
- reply that page is cached only for GET request (should probably be HEAD too actually)
- page must not contains form (due to csrf tokens, but actually this should be removed due to next point)
- it actually does db lookup and generate the page, then hash it, so it saves only bandwidth, not cpu/io

This should be removed when more intelligent caching get implemented

igalic:
can you propose a patch out of this situation?

and do you believe it would help us with our current performance issues, before we switch to async rocket?

trinity-1686a:
I'll look into it if I got some time.

[...]

PR #794, opened 2020-06-25
Better caching

elegaanz:
Based on @trinity-1686a's patch.

I'm not sure it is a fix for #786 but it is better than nothing.


elegaanz:
OK, my git history is broken, I need to fix that.

src/template_utils.rs
```diff
@@ -63,7 +63,7 @@ impl<'r> Responder<'r> for Ructe {
         let etag = format!("{:x}", hasher.finish());
         if r.headers()
             .get("If-None-Match")
-            .any(|s| s[1..s.len() - 1] == etag)
+            .any(|s| s[1..s.len() - 1] == etag || s[3..s.len() - 1] == etag)
```
igalic, line 66: can one of you two explain what this does?

igalic:
don't we still have to to change the subscribe/follow/blah button from a form… to… well… a button?

src/template_utils.rs
trinity-1686a: this should probably get it's comment actually.
We send Etags as `'some_tag'`, the old match does not verify the `'`, but check that whatever in-between is the right tag. Nginx apparently change the Etag to `W/'some_tag'`, W/ indicating it's a weak Etag, the content may not be byte-equal, but is semantically identical (which is imo a shitty but correct interpretation of the RFC, as Nginx send compressed data)

The new match ignore that W/ saying it's a weak etag and so match if Nginx modified the header we sent previously

Comment might looks like : // if a strong or weak etag matches

trinity-1686a:
well there is that too, but these buttons are not here on the front page or a blog or user page. Anyway for post a better caching solution that does not require to query the database should be added

FreyjaWildes (approved):
I'm not totally sure, being new to the software and Rust, but this seems fair enough and a small change. Probably can be approved, right ?

KitaitiMakoto:
@FreyjaWildes You're right. I merged this pull request as https://git.joinplu.me/Plume/Plume/pulls/840

Thank you for keeping interest on Plume!

Issue #807, opened 2020-08-03
menu button still not working on ios 13.5.1


cybergipsy2020:
I noticed that the menu button on the top left corner still won't respond on ios browsers. 

- **Plume version:** You can find it in the footer of your instance. If you know the exact commit, please also add it.
0.4.0
- **Operating system:**
ios 13.5.1
- **Web Browser:**
safari, firefox

marek-lach:
I believe this would be remedied by merging #663 @elegaanz, @igalic?

igalic:
https://github.com/Plume-org/Plume/pull/663#issuecomment-530256796 is this fixed though?

marek-lach:
> [#663 (comment)](https://github.com/Plume-org/Plume/pull/663#issuecomment-530256796) is this fixed though?

Haven't tested it live myself, but it should definitely bring some improvements, perhaps in combination with #810...

PR #810, opened 2020-08-04
Improve burger menu performance with Webkit on mobile

marek-lach:
Attempt to fix the menu in Safari on mobile iOS.

Should address #807, at least in part.

### DIFF
--- a/assets/themes/default/_header.scss
+++ b/assets/themes/default/_header.scss
@@ -19,7 +19,6 @@ body > header {
 
     a {
       transform: skewX(15deg);
-      -webkit-transform: translateZ(0);
      	display: flex;
      	flex-direction: column;
      	align-items: center;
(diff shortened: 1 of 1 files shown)


elegaanz:
Sorry this PR was closed, it is probably a bug with mirroring from Gitea. Could you please push your branch to https://git.joinplu.me and re-open the PR here?

Also, about the code itself: isn't `transform` supported everywhere? According to [this table](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#Browser_compatibility) it should work on Safari since iOS 9, without the `-webkit-` prefix

marek-lach:
> about the code itself: isn't `transform` supported everywhere? According to [this table](https://developer.mozilla.org/en-US/docs/Web/CSS/transform#Browser_compatibility) it should work on Safari since iOS 9, without the `-webkit-` prefix

Webkit browsers, especially iOS, would benefit most visibly from a simple GPU acceleration of the menu animation.

Firefox seems to work just fine with the more specific `transform` values already given, so does not need fixing.

I know you don't like browser specific hacks, so it's best to keep them to a minimum, but sometimes they do unfortunately still seem necessary...

igalic:
migrated over to https://git.joinplu.me/Plume/Plume/pulls/820

Issue #754, opened 2020-04-18
Show suitable error instead of not found page for duplicated blocklist emails


KitaitiMakoto:
Hello,

When you attempt to add an email to blocklist and it already exists in blocklist, you see not found page. But you should see error "an email you're adding already exists in blocklist". Or, just ignoring error and showing blocklist page might be ok.

### How to reproduce:

1. Add an email(say, `invalid@example.com` here) to blocklist at `/admin/emails`
2. Again, add the same email(`invalid@example.com`) to blocklist
3. You see an error message "We couldn't find this page."

### Workaround:

Just use browser back button.

I know you, Plume developers, has too little time to do anything, and I might fix this bug. But, sorry, I want to spend my time to other things like creating custom theme and UI translation. So, I create this issue to prevent us forgetting this problem and show a workaround to people who encounter the same problem.

- **Plume version:** You can find it in the footer of your instance. If you know the exact commit, please also add it.
  - c217e5e
- **Operating system:**
  - Ubuntu 18.04 on Windows 10 WSL
- **Web Browser:**
  - Firefox

KitaitiMakoto:
Fixed by #818

PR #818, opened 2020-10-06
Fix displaying not found page when submitting a duplicated blocklist email

serra-allgood:
This PR addresses issue #754. I'm still new to Rust and even newer to Plume, but I thought this approach made sense. The alternative I thought of was to leverage the `BlocklistedEmail::matches_blocklist` function, but I thought that approach is unreasonable because it requires loading the entire blocklisted_emails table.


igalic (approved):
🤔 this looks reasonable

KitaitiMakoto:
I think so, too.

elegaanz:
For future contributions, we moved to [a self-hosted gitea](https://git.joinplu.me) so we prefer PRs here even if we can merge them "manually" from Github. Anyway, thank you for your help :)

PR #832, opened 2020-12-23
Change riot.im link to matrix.to

resynth1943:

### DIFF
--- a/templates/base.rs.html
+++ b/templates/base.rs.html
@@ -93,7 +93,7 @@
                 <h3>@concat!("Plume ", env!("CARGO_PKG_VERSION"))</h3>
                 <a href="https://docs.joinplu.me">@i18n!(ctx.1, "Documentation")</a>
                 <a href="https://github.com/Plume-org/Plume">@i18n!(ctx.1, "Source code")</a>
-                <a href="https://riot.im/app/#/room/#plume:disroot.org">@i18n!(ctx.1, "Matrix room")</a>
+                <a href="https://matrix.to/#/#plume:disroot.org">@i18n!(ctx.1, "Matrix room")</a>
             </div>
         </footer>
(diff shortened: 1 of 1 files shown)


templates/base.rs.html
```diff
@@ -93,7 +93,7 @@ <h3>@Instance::get_local().map(|i| i.name).unwrap_or_default()</h3>
                 <h3>@concat!("Plume ", env!("CARGO_PKG_VERSION"))</h3>
                 <a href="https://docs.joinplu.me">@i18n!(ctx.1, "Documentation")</a>
                 <a href="https://github.com/Plume-org/Plume">@i18n!(ctx.1, "Source code")</a>
-                <a href="https://riot.im/app/#/room/#plume:disroot.org">@i18n!(ctx.1, "Matrix room")</a>
+                <a href="https://matrix.to/#/#plume:disroot.org">@i18n!(ctx.1, "Matrix room")</a>
```
igalic, line 96: we should also move from distroot.io to… something that exists…

igalic (commented):
🙋🏻‍♀️

templates/base.rs.html
elegaanz: This is just a room alias, matrix rooms are not really linked to a homeserver I think, so there is no need to create a new room, just to add a new alias (and I believe there already are some other aliases for the Plume rooms).

resynth1943: Send me another address, and I'll add it in.

KitaitiMakoto: You may use `#plume-blog:matrix.org` as an alias.

resynth1943: See 07a6be4.

igalic (approved):

KitaitiMakoto:
Thank you!

PR #856, opened 2021-09-05
Make Feature list more clear

dhuschde:
Say it, when something is currently missing.
Otherwise, people will get confused.


KitaitiMakoto:
Good catch!

PR #1129, opened 2026-02-25
Upgrade dependencies

Timshel:
Hey

Was interested in the project but when trying to build it failed.
So I spend some time upgrading Rocket and Diesel dependencies and it's now compiling.

Has not tested it much for now, just created an account uploaded an avatar and created an article.

Just wanted to open a PR to get some feedbacks (Will open on Forgejo if there is some :)


elegaanz:
There is a Forgejo here: https://git.joinplu.me/Plume/Plume

But the project is more or less unmaintained.

